PR #9, opened 2017-12-14
remove tox from requirements

### LINKED ISSUE #6: separate requirements
Seems useful for test/dev but we'll never use it for prod. Separate requirements and make prod as slim possible.

https://github.com/uisautomation/sms-webapp/blob/d63400151b0d22bf552d7c1e795e1acff4238d38/requirements.txt#L6

https://github.com/uisautomation/sms-webapp/blob/d63400151b0d22bf552d7c1e795e1acff4238d38/requirements.txt#L9

rjw57:
closes #6

### DIFF
--- a/doc/developer.rst
+++ b/doc/developer.rst
@@ -37,8 +37,7 @@ build only the documentation:
 
     The CircleCI job runs tox configured with a PostgreSQL database to match
     that deployed in production. To replicate this, either run a PostgreSQL
-    instance on the local machine, use the Google Cloud SQL proxy or :any:`run
-    the tests via docker-compose <docker-tox>`.
+    instance on the local machine or use the Google Cloud SQL proxy.
 
 .. _devserver:
 
@@ -109,25 +108,6 @@ can launch such a shell via:
 .. code-block:: bash
 
     $ docker-compose run shell
-
-.. _docker-tox:
-
-Running tests within the container
-``````````````````````````````````
-
-The test-suite can be run within the container using a PostgreSQL database in
-the following way:
-
-.. code-block:: bash
-
-    $ docker-compose tox
-
-.. note::
-
-    Unlike the other docker-compose services, the tox service *does not* mount
-    the local repository as a volume. This is because tox needs to write to the
-    filesystem. Make sure that the container build is up to date via
-    ``docker-compose build`` before running tests.
 
 Cloud infrastructure
 --------------------
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,9 +29,6 @@ services:
       - DJANGO_DB_PASSWORD=databasePass
       - DJANGO_EMAIL_HOST=debugsmtp
       - DJANGO_EMAIL_PORT=1025
-  tox:
-    <<: *devserver
-    entrypoint: ["tox"]
   migrate:
     <<: *devserver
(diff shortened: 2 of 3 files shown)


abrahammartin:
so, the way I understand it, tox was used to execute the test suite inside the container, no? I was suggesting to remove tox from requirements.txt and put in in something like requirements_dev.txt but you have removed it completely. I understand that you think this is now not necessary at all (as removed completely from code) as we have the CircleCI executing tests for each commit. Is my understanding correct?

rjw57:
Sort of. We already have multiple ways of running tox: we can run it directly or via CircleCI. The advantage of having it run via docker-compose is that a local developer can run the tests using Postgres via tox. A developer could already do this via ``docker-compose run manage test`` so that use was marginal.

To add the ability back in, the container would have to install test suite requirements which raises the question: if our container has to install package X to function for our needs, why then is X not a general requirement?

rjw57:
> To add the ability back in, the container would have to install test suite requirements which raises the question: if our container has to install package X to function for our needs, why then is X not a general requirement?

This of course begs the question of whether we want the container built by the Dockerfile we ship to be suitable for production use.

rjw57:
> To add the ability back in, the container would have to install test suite requirements

Actually, that's not true. If we wired up ``setup.py test`` to run tests then we could add ``tox`` to ``tests_require`` in ``setup()``.

abrahammartin:
> Actually, that's not true. If we wired up setup.py test to run tests then we could add tox to tests_require in setup().
I personally think is better to keep the production environment as clean as possible (of requirements/dependencies/external packages/etc that are not directly required to run the prod environment). So for example, django_debug_toolbar is also very useful and I sometimes include it in the dev requirements, but obviously I don't want to include that in the general requirements.txt (for production as well). What do you think?

rjw57:
That makes some sense but then we'd have to think carefully about what goes in this container. Is this container the production one, or one used only for local development?

Having peered at the GCE docs, I'm more convinced than not that we should have all the deployment stuff in a separate repo and have a Dockerfile *there* which checks out this repo and builds a specialised production container. (Like the way the official Python image downloads the Python source rather than using COPY.)

This has the following advantages as I see it:

1. The container can be tagged with a git commit hash and we know absolutely that that container corresponds to that hash. (I believe this is important when we come to deploy since kubernetes rolling deployment upgrades only happen if the image name/tag changes?)
2. Avoiding COPY/ADD means we do not risk accidentally copying a file into the container we do not intend.
3. We can separate deployment configuration from this repo which keeps us honest w.r.t. making this repo "deployment neutral".

So, in summary, I'm edging towards making the Dockerfile/docker-compose.yml we ship in this repo be developer-only. Thoughts?

rjw57:
> I'm edging towards making the Dockerfile/docker-compose.yml we ship in this repo be developer-only.

By which I mean that it also ``pip install``-s a custom requirements.txt file which contains things like tox, django_debug_toolbar, etc, etc.

abrahammartin:
@rjw57 and I agreed that the repo should contain dev/test requirements/Dockerfile/etc and set up a new repo that forks this one to contain production requirements/Dockerfile/docker-compose/etc

rjw57:
OK, will close this PR and merge it into #11 ?

PR #11, opened 2017-12-14
rewrite the bundled Dockerfile to be based on official Python 3.6 image

### LINKED ISSUE #7: Google Engine
https://github.com/uisautomation/sms-webapp/blob/d63400151b0d22bf552d7c1e795e1acff4238d38/Dockerfile#L21

After reading https://github.com/GoogleCloudPlatform/python-runtime and https://cloud.google.com/appengine/docs/flexible/python/configuring-your-app-with-app-yaml I can't see any advantage of using Google's image.

Any convincing argument?

rjw57:
Instead of basing ourselves on the GCE base image for Python, base
ourselves on the upstream Python image.

Closes #7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,32 +1,26 @@
-#	Copyright 2015, Google, Inc.
-#	Copyright 2017, University of Cambridge
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


rjw57:
Closing since in #6 we decided the Docker container would be devel-only.

rjw57:
Replaced by #14

PR #13, opened 2017-12-14
Add Flake8 support for tox test runner

### LINKED ISSUE #12: Add code-style checking to circle-ci config
Should we add ``flake8``-style checking to the circleci config to require PEP 8 compliance? (Or that non-compliance is tagged with a comment.)

rjw57:
Add flake8 as a code-style check to tox test runner. Since CircleCI runs the default tox envs, this means that code-style should be enforced for each pull request.

Closes #12


rjw57:
I hadn't noticed that I have some squashed flake8 errors set globally to reduce noise with our other projects. The project should now be flake8 clean.

.editorconfig
```diff
@@ -1,12 +1,8 @@
...
 
-[*.ini]
-indent_style=space
-indent_size=4
-
-[*.{yml,yaml}]
-indent_style=space
-indent_size=2
+[*.py]
+max_line_length=79
```
abrahammartin: I'd vote for 119 line length. @msb ? @rjw57 ?

rjw57: We would have to add configuration to flake8 to ignore line length since it enforces PEP8: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

rjw57: And I don't know off the top of my head if flake8 supports a configurable line length as opposed to just ignoring line length errors.

rjw57: And even if we make use of PEP8's wiggle room, to paraphrase Monty Python and the Holy Grail "99 is the maximum line length and the length of the maximum line shall be 99. 112 is right out!"

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters.

rjw57: The consensus appears to be "99".

PR #14, opened 2017-12-14
add development Docker image

rjw57:
Instead of basing ourselves on the GCE base image for Python, base
ourselves on the upstream Python image.

Make the image explicitly developer focussed by installing
developer-friendly packages and using the development server by default.

Add in the django debug toolbar since we are now free to install
developer-only modules.


rjw57:
Resolved in 8bccaed

PR #17, opened 2018-01-02
add placeholder empty response for the index

rjw57:
Add a temporary view to the smswebapp *project* which renders an empty
HTML page and wire it up to the index view. This is required since the
Google Load Balancers use an HTTP end point as a health monitor[1].
Since all of our views currently re-direct to a login page or return
404, there wasn't a simple "will succeed" endpoint.

In the fullness of time, we could move this to "/health-monitor" or
"/status" if we ave something useful to put there.

[1] https://cloud.google.com/compute/docs/load-balancing/http/


abrahammartin:
You could use the status module in django-automationcommons that SSGW and Research Dashboard use: https://git.uis.cam.ac.uk/i/uis/automation/django-automationcommon.git/blob/HEAD:/automationcommon/views.py

It is an attempt to standarise our status pages on all our services

rjw57:
Will re-factor this PR to use the status view...

rjw57:
> Will re-factor this PR to use the status view...

And, done.

PR #19, opened 2018-01-03
remove circleci configuration

rjw57:
If Travis can run tests using postgres, CircleCI lost its USP.


rjw57:
Yes, this branch has no travis configuration (so it fails) and removes the CircleCI configuration (so it fails)

PR #23, opened 2018-01-05
doc: add "githubpages" sphinx extension

rjw57:
GitHub pages uses jekyll as a static site builder which doesn't
interact well with directory names generated by Sphinx such as
"_static". Add the githubpages Sphinx extension which adds a ".nojekyll"
file to the output which stops GitHub trying to compile our static site.

See also:
https://help.github.com/articles/files-that-start-with-an-underscore-are-missing/


rjw57:
The result of this PR is currently deployed to https://rjw57.github.io/sms-webapp/.

PR #24, opened 2018-01-05
Documentation fixes and clarifications

rjw57:
Various clarifications to the documentation. Aside from fixing some errors of markup, document the correct method of running tox via docker-compose, remove the redundant Cloud deployment notes and document the base settings.


docker-compose.yml
```diff
@@ -29,10 +29,21 @@ services:
...
       - DJANGO_EMAIL_HOST=debugsmtp
       - DJANGO_EMAIL_PORT=1025
+      - JWPLATFORM_API_KEY=${JWPLATFORM_API_KEY}
+      - JWPLATFORM_API_SECRET=${JWPLATFORM_API_SECRET}
+      - JWPLATFORM_EMBED_PLAYER_KEY=${JWPLATFORM_EMBED_PLAYER_KEY}
   tox:
     <<: *devserver
     entrypoint: ["tox"]
     command: [""]
+    # DANGER, WILL ROBINSON! Making the application non-read only means that tox
```
abrahammartin: Set up toxworkdir to avoid that?

rjw57: So, after some discussion, we decided that the dangers of having random root-owned files appears outweighs the benefits. Proposed solution: have toxworkdir and build directory in some well-known temporary location(s) and if one really, really cares about the output, one can mount a volume.

PR #32, opened 2018-01-15
Make it fail

abrahammartin:


smswebapp/test/test_admin.py
```diff
@@ -28,4 +28,4 @@ def test_authenticated(self):
         """Authenticated log in to admin as superuser succeeds."""
         self.client.force_login(self.superuser)
         r = self.client.get(reverse('admin:index'))
-        self.assertEqual(r.status_code, 200)
+        self.assertEqual(r.status_code, 400)
```
rjw57: IS this RFC 2616 compliant?

Issue #38, opened 2018-01-30
RSS feeds


abrahammartin:
The new sms web app needs to generate the same RSS feeds as the old ones do for media items and collections. Examples:

- http://rss.sms.cam.ac.uk/rss/media/2657898
- http://rss.sms.cam.ac.uk/rss/collection/2448176

Once this story is completed, we will move rss.sms.cam.ac.uk from the old sms application to the new sms application

abrahammartin:
duplicated from #76

PR #49, opened 2018-04-09
Add RSS endpoint for media items

rjw57:
JWPlatform supports directly rendering an (m)RSS feed for a given item by means of its platform delivery API. This API has a (third) method for authentication so, firstly, 0a2528f implements that method anf then 52b16cc uses it to add the endpoint. The endpoint simply redirects to the (signed) URL.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ omit=
 	setup.py
 	manage.py
 	*/test/*
+	*/tests/*
 	smswebapp/settings/*
 	smswebapp/wsgi.py
 	.tox/*
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,6 +14,7 @@ requests-oauthlib
 
 # To interact with the jwplatform API
 jwplatform
+pyjwt
 
 # For an improved ./manage.py shell experience
 ipython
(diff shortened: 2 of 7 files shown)


abrahammartin:
The code is correct but I don't think this closes the issue completely as we need to generate collection RSS feeds as well using the collection_id from the video tags, see https://sms.cam.ac.uk/collection/2448176 and http://rss.sms.cam.ac.uk/rss/collection/2448176

rjw57:
The example given in the issue was for media items only. I thought the story was scoped to them. If not, it needs resizing.

abrahammartin:
I have removed the Closes ... so it can be merged now without closing the ticket, although I let you decide if you want to merge or not given the feedback form codecov

abrahammartin:
#38 goal is to replace RSS feeds, all of them, and replace the endpoint with apache to use the new ones (like with the video iframe) otherwise we can't move to jwplayer as the RSS feeds will point to the old SMS media item and it won't work.

rjw57:
I've updated the description of #38. Does that cover all feeds?

Re coverage: I was a fool and didn't check in the tests(!)

PR #52, opened 2018-04-20
Redirect back to legacy SMS is a media item cannot be found

### LINKED ISSUE #51: Redirect to old SMS if video not found on JWPlatform
In all legacy views, if a video cannot be found on JWPlatform, we should redirect back to the old SMS (maybe via Apache magic) in order to make sure that we hide the latency between SMS processing finishing and us syncing to JWPlatform.

rjw57:
> **IMPORTANT:** This PR includes the flake8 fixes from #53 so review that first.

This PR addresses #51 but takes the opportunity to do a bit of housekeeping. Specifically, we move all the code which interacts with the legacy SMS system to its own Django application. The ideal would be, therefore, that at some point we can remove the application from ``INSTALLED_APPS`` and carry on with little disruption.

This PR is best reviewed commit-wise. Since GitHub likes to mess with the order of commits, they should be reviewed in the following order:

* 938ca24 simply fixes some PEP8 whitespace problems.
* 8d55c37 adds a blank ``legacysms`` application based on the one from uisautomation/django-boilerplate#22.
* 8af242c moves the legacy SMS-related views into the ``legacysms`` application. Functionality related to the JWPlatform API remains where it was since it will be of use to the application implementing the new SMS.
* 43786eb introduces the principle that the shape of URLs we support in ``legacysms`` should match those of the SMS to make it easier to do redirection.
* 4a3fccc removes the unused URL routing section from the documentation for ``legacysms``
* 7b51c8a implements a ``legacysms.redirect`` module which can generate redirect responses for passing results back to the legacy SMS.
* fae5a87 makes use of the ``legacysms.redirect`` module to replace 404 errors with redirects.
* 0361d8d adds an example Apache configuration which can be used on the legacy SMS to interact with this web application.

If merged, the configuration similar to that in 0361d8d can be added to the legacy SMS to off-load media embedding and RSS feed generation to this web application.

Closes #51


rjw57:
@abrahammartin code review done

PR #54, opened 2018-04-20
enable flake8 testing

### LINKED ISSUE #53: flake8 is not running correctly
The tox.ini file lists only ``smswebapp`` and ``doc`` as those directories scanned by flake8. This should be replaced with ``.``.

rjw57:
As noted in #53, the PEP8 tests were only running on a subset of the code. This is fixed by d016a6a. The preceding commit, 7d216dc, fixes the issues which were not previously being caught.

Closes #53.

### DIFF
--- a/smsjwplatform/admin.py
+++ b/smsjwplatform/admin.py
@@ -1,3 +1,3 @@
-from django.contrib import admin
+# from django.contrib import admin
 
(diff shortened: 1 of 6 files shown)


abrahammartin:
tested `docker-compose up tox` with no flake8 errors

PR #55, opened 2018-04-20
Add media download endpoint

### LINKED ISSUE #50: Add endpoint for direct download of media
The existing videos can be directly downloaded via, e.g., the "Available formats" links on https://sms.cam.ac.uk/media/2724577. We need a matching endpoint under /legacy.

rjw57:
> **IMPORTANT:** This PR makes use of functionality from #52 and #54 so review those first.

This PR adds a new endpoint to support the legacy SMS's media download URLs. Since this endpoint makes heavy usage of logging to explain its flow, 47dcd5a takes the opportunity to increase the default chatty-ness of logging so that the logs will be visible when the application is deployed.

579e915 adds logic to redirect back to the SMS if we can't handle a download request. The existing redirects can directly be handled by the SMS frontend web application but downloads are served by a separate machine. Rather than requiring the configuration of two base URLs for redirection, we require that the web application be configured to handle download URLs as well. The example Apache configuration in the documentation is updated to include this.

956330f implements the download endpoint itself. Some heuristics are used to match the requested URL to the appropriate JWPlatform source. If we cannot find the matching file, we use the redirect functionality from 579e915 to redirect back to the legacy SMS application.

Closes #50


rjw57:
@abrahammartin  I think I should've addressed all your points. I'll tweak the configs for sms-dev to understand the _legacy magic.

PR #10, opened 2017-12-14
add a basic ucamwebauth configuration

rjw57:
Along with the configuration, we test that the admin actually requires login, that logging in via tests work with the newer force_login() method on the test client and we remove the confusing old GCloud stuff.

Closes #5

PR #56, opened 2018-04-24
add basic Docker-isation

rjw57:
Add a simple Dockerfile for the smswebapp which can be used to host the app. Take the opportunity to rationalise the requirements files into a single directory to avoid proliferation as outlined in uisautomation/lookupproxy#10.


Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM python:3.6-alpine
```
abrahammartin: should we create a docker-compose for the production Dockerfile as well?

rjw57: Good idea. Inspired by the [standard cookiecutter Django template](https://github.com/pydanny/cookiecutter-django/tree/master/%7B%7Bcookiecutter.project_slug%7D%7D) I think this is the threshold where we split them into separate files.

rjw57:
OK, split the docker-compose and took the opportunity to improve the documentation around the docker-compose stuff. If this gets merged, I'll teleport it to django-boilerplate.

compose/development.Dockerfile
```diff
@@ -1,10 +1,14 @@
-FROM python:3.6
+FROM python:3.6-alpine
```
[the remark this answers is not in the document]
rjw57: I've no strong opinion on this so I'm happy to make this ``FROM python:3.6``.

rjw57:
@abrahammartin Code review done

compose/tox.yml
```diff
@@ -8,7 +8,7 @@ services:
   tox:
     build:
       context: ..
-      dockerfile: ./Dockerfile
+      dockerfile: ./compose/development.Dockerfile
```
abrahammartin: isn't it better to execute tox using the production enviroment (alpine) rather that the dev enviorment? I was only thinking on the dev enviroment (debian) for devserver / debug puproses

rjw57: Ah, OK. Yes, we can do that. Will fix.

rjw57: And, done in ea2a759

rjw57:
Merged after @abrahammartin indicated at this morning's standup that it can be moved to "done".

PR #57, opened 2018-04-25
moved oauth related utilities to django-oauthcommon

msb:


smsjwplatform/defaultsettings.py
```diff
@@ -44,34 +44,34 @@
 
 """
 
-SMS_OAUTH2_CLIENT_ID = None
```
rjw57: These default settings hsould be moved to automationoauthclient or automationoauthdrf as appropriate

PR #62, opened 2018-05-25
New production dockerfile

abrahammartin:
The new production dockerfile moves to use a script as an entrypoint that will execute migrations before starting gunicorn.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,13 +28,5 @@ ENV \
 	DJANGO_SETTINGS_MODULE=smswebapp.settings.docker \
 	PORT=8000
 
-# Use gunicorn as a web-server after running migration command
-CMD gunicorn \
-	--name smswebapp \
-	--bind :$PORT \
-	--workers 3 \
-	--log-level=info \
-	--log-file=- \
-	--access-logfile=- \
-	--capture-output \
-	smswebapp.wsgi
+# Use gunicorn as a web-server
+ENTRYPOINT ["/usr/src/app/docker-entrypoint.sh"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -28,13 +28,5 @@ ENV \
...
-	--name smswebapp \
-	--bind :$PORT \
-	--workers 3 \
-	--log-level=info \
-	--log-file=- \
-	--access-logfile=- \
-	--capture-output \
-	smswebapp.wsgi
+# Use gunicorn as a web-server
+ENTRYPOINT ["/usr/src/app/docker-entrypoint.sh"]
```
rjw57: After some discussion, we agreed that we'd try running the migration as an [Init Container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/). With that in mind, back out the ``docker-entrypoint.sh`` addition and modification to ``compose/tox.yml``.

.circleci/config.yml
```diff
@@ -39,10 +39,12 @@ jobs:
           command: |
             docker-compose --file compose/base.yml --file compose/production.yml up -d
             ./compose/wait-for-it.sh localhost:8000 -t 15
-            docker-compose --file compose/base.yml --file compose/production.yml exec production_app ./manage.py migrate
+      # Despite the webserver working on 8000, the web app still requires a few secs until is ready to show /healthz
```
rjw57: Could we handle this in a more elegant fashion by running ``curl`` in a loop with a ``sleep 1`` and early-outing if we succeed?

PR #78, opened 2018-06-27
Change uses of JWP management API to use CachedResource

### LINKED ISSUE #68: Update backend to use the new local database instead of JWPlatform APIs
Use the new django models to search for the media items.

rjw57:
This PR changes two uses of the JWP management API which were on the critical path of serving resources to users to make use of the CachedResource database objects.

Closes #68.


abrahammartin:
CircleCI failed

abrahammartin:
Apparently the reason why it's failing is CircleCI npm's is down: https://status.circleci.com/

Issue #39, opened 2018-01-30
Individual media pages


abrahammartin:
Develop an equivalent media page in the new sms app with the same features but using the new material design style.

rjw57:
Needs some discussion and perhaps splitting. E.g. do we want to structure this as a very thin API proxy to JWPlatform and then write much of the frontend in React?

PR #92, opened 2018-06-28
Individual media pages

msb:
Implements a simple media page with working player and a source download link. An outstanding issue is that localhost:3000 won't work correctly for the new page (still works for the home page).

closes: #39 
closes: #95


frontend/src/pages/MediaPage.js
```diff
@@ -0,0 +1,106 @@
...
+    <div className={ classes.body }>
+      <section>
+        <Grid container spacing={16} className={ classes.gridContainer }>
+          <Grid item xs={12} className={ classes.playerWrapper } style={{paddingBottom:'56.25%'}}>
+            <iframe src={mediaItem.player_url} className={ classes.player } width="100%" height="100%" frameBorder="0" allowFullScreen>
+            </iframe>
+          </Grid>
+          <Grid item xs={12}>
+            <Typography variant="subheading" gutterBottom>
+              <a target='_blank' href={mediaItem.sources[0].url}>
```
rjw57: ``mediaItem.bestSource``?

frontend/src/pages/MediaPage.js
```diff
@@ -18,6 +19,8 @@ const MediaPage = ({ mediaItem, classes }) => (
       <ProfileButton variant="flat" color="inherit" />
     </AppBar>
     <div className={ classes.body }>
+      <MotdBanner />
```
rjw57: I think this is a shibboleth that we should refactor this into a general "Page" component.

api/serializers.py
```diff
@@ -23,13 +36,34 @@ class MediaSerializer(serializers.Serializer):
...
         help_text='A URL for the media item. This is a URL for the media UI, not a resource URL.'
     )
+    player_url = serializers.SerializerMethodField(
+        help_text='A URL to retrieve an embeddable player for the media item.'
+    )
+    sources = SourceSerializer(
+        help_text='A collection of download URLs for different media types.',
+        required=False, many=True
+    )
+    legacy_stats_url = serializers.SerializerMethodField(
```
rjw57: I think we should avoid having the legacy SMS mentioned explicitly. Instead, just have ``stats_ui_url``. Then when we add a native stats page, the API doesn't need to change.

api/serializers.py
```diff
@@ -23,13 +36,34 @@ class MediaSerializer(serializers.Serializer):
...
+        help_text='A URL linking to the stats page in the legacy SMS app for the media item.'
+    )
 
     def get_ui_url(self, obj):
-        return 'https://sms.cam.ac.uk/media/{.media_id}'.format(obj)
+        return '/media/{[key]}'.format(obj)
```
rjw57: Can we use Django's ``reverse`` helper here?

api/tests/test_views.py
```diff
@@ -129,17 +130,70 @@ def test_basic_list(self):
...
+
+        self.assertEqual(response.data['id'], 'mock1')
+        self.assertEqual(response.data['title'], 'Mock 1')
+        self.assertEqual(response.data['description'], 'Description for mock 1')
+        self.assertEqual(response.data['published_at_timestamp'], 1234567)
+        self.assertEqual(
+            response.data['poster_image_url'], 'https://cdn.jwplayer.com/thumbs/mock1-720.jpg'
+        )
+        self.assertEqual(response.data['duration'], 54.0)
+        self.assertEqual(response.data['ui_url'], '/media/mock1')
```
rjw57: Can we test against what Django's ``reverse`` helper returns? Then is we change the URL, we don't need to change all of the tests.

api/urls.py
```diff
@@ -27,6 +27,7 @@
 urlpatterns = [
     path('collections/', views.CollectionListView.as_view(), name='collections'),
     path('media/', views.MediaListView.as_view(), name='media'),
+    path('media/<media_key>', views.MediaView.as_view(), name='media'),
```
rjw57: This path has the same name as the one above it. (Goddamned words which don't change in the plural.) Perhaps rename the views to ``media_list`` and ``media_item``?

frontend/src/components/AppBar.js
```diff
@@ -2,14 +2,16 @@ import React from 'react';
...
-import Button from '@material-ui/core/Button';
 import Grid from '@material-ui/core/Grid';
 import Toolbar from '@material-ui/core/Toolbar';
 import Typography from '@material-ui/core/Typography';
 import { withStyles } from '@material-ui/core/styles';
 
 import SearchForm from './SearchForm';
 
+// The location for a redirected search request
+const SEARCH_LOCATION = '/';
```
rjw57: When we sort out routing in #100 or #97, I think we'll need to re-factor this into a separate ``/search`` route and turn the AppBar's search box into something which re-directs. I've opened #102 to track this.

frontend/src/components/AppBar.js
```diff
@@ -126,8 +136,13 @@ const handleSubmit = (event, onSearch) => {
...
 
-  // Pass query to handler.
-  onSearch(query);
-}
+  if(onSearch) {
+    // Pass query to handler.
+    onSearch(query);
+  } else {
+    // no search handler so redirect the search
+    location = SEARCH_LOCATION + '?q=' + encodeURI(query);
```
rjw57: See #102

frontend/src/components/ProfileButton.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  return (
+    <Button {...otherProps}>
+      { profile.username }
+    </Button>
+  );
+};
+
+const ProfileButtonWithProfile = withProfile(ProfileButton);
```
rjw57: I think this should be in a separate module, probably under ``src/containers``. From the [learn react blog](https://medium.com/@learnreact/container-components-c0e67432e005):

> A container does data fetching and then renders its corresponding sub-component. That’s it.

So there'd be a module, ``src/containers/ProfileButtonContainer.js`` with a correspondingly named component, ``ProfileButtonContainer`` which is simply ``ProfileButton`` wrapped with ``withProfile``.

frontend/src/index.tsx
```diff
@@ -2,5 +2,18 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 
 import IndexPage from './pages/IndexPage';
+import MediaPage from './pages/MediaPage';
 
-ReactDOM.render(<IndexPage />, document.getElementById('root') as HTMLElement);
+const mappings = {
+  'media': <MediaPage/>,
+  'root': <IndexPage/>,
+};
```
rjw57: Again, with #102, we'd want to replace this with some router.

frontend/src/pages/MediaPage.js
```diff
@@ -0,0 +1,89 @@
...
+        <Grid item xs={12} className={ classes.playerWrapper } style={{paddingBottom:'56.25%'}}>
+          <iframe src={mediaItem.player_url} className={ classes.player } width="100%" height="100%" frameBorder="0" allowFullScreen>
+          </iframe>
+        </Grid>
+        <Grid item xs={12}>
+          <Typography variant="subheading" gutterBottom>
+            <a target='_blank' className={ classes.link } href={mediaItem.bestSource.url} download>
+              Download media
+            </a>
+            <a className={ classes.link } href={mediaItem.legacy_stats_url} style={{float: 'right'}}>
```
rjw57: Rather than having floats, could we put a nested grid?
```jsx
<Grid item xs={12}>
    <Grid container>
        <Grid item xs={6}>/* Download link */</Grid>
        <Grid item xs={6}>/* Stats link */</Grid>
    </Grid>
</Grid>
```

smsjwplatform/jwplatform.py
```diff
@@ -198,6 +211,80 @@ def from_media_id(cls, media_id, preferred_media_type='video'):
...
+        field = parse_custom_field('acl', self.get('sms_acl', 'acl:WORLD:'))
+
+        # Work around odd ACL entries. See uisautomation/sms2jwplayer#30.
+        if field == "['']":
+            return []
+
+        return [acl.strip() for acl in field.split(',') if acl.strip() != '']
+
+    @classmethod
+    def from_key(cls, key, client=None):
```
rjw57: ``client`` is unused here? Perhaps replace with ``session=None``? (See below.)

smsjwplatform/jwplatform.py
```diff
@@ -198,6 +211,80 @@ def from_media_id(cls, media_id, preferred_media_type='video'):
...
+
+        :param key: JWPlatform key for the media.
+        :param client: (optional) an authenticated JWPlatform client as returned by
+            :py:func:`.get_jwplatform_client`. If ``None``, call :py:func:`.get_jwplatform_client`.
+
+        :raises: :py:exc:`VideoNotFoundError` if the video is not found.
+
+        """
+        # Fetch the media download information from JWPlatform.
+        response = DEFAULT_REQUESTS_SESSION.get(
```
rjw57: It strikes me that "default" here means "the only one you can use". Perhaps replace ``client`` in the arguments with ``session`` and have:
```python
session = session if session is not None else DEFAULT_REQUESTS_SESSION
```
at the top of the function and have this line be
```python
response = session.get( ... )
```

smsjwplatform/urls.py
```diff
@@ -2,18 +2,15 @@
 Default URL patterns for the :py:mod:`smsjwplatform` application are provided by the
 :py:mod:`.urls` module. You can use the default mapping by adding the following to your global
 ``urlpatterns``:
-
 .. code::
-
```
rjw57: This blank line is required by the RST specification to separate options from content.

smsjwplatform/urls.py
```diff
@@ -2,18 +2,15 @@
...
     from django.urls import path, include
-
```
rjw57: I think this blank line adds clarity and is consistent with PEP8 spacing.

smswebapp/settings/base.py
```diff
@@ -220,6 +221,9 @@
     os.path.abspath(os.path.join(BASE_DIR, 'frontend', 'build'))
 )
 
+# The base url for Media API - used by the 'ui' application
+MEDIA_API_URL = 'http://localhost:8080/api/'
```
rjw57: Why not ``/api``?

.circleci/config.yml
```diff
@@ -51,16 +51,13 @@ jobs:
...
-
-      - run:
-          name: Push docker image
+          name: Tag & Push docker image
           command: |
             if [ ! -z "$GCLOUD_SERVICE_KEY" ]; then
+              docker tag production:latest eu.gcr.io/$GCLOUD_PROJECT/$CIRCLE_PROJECT_REPONAME:sha-$CIRCLE_SHA1
+              docker tag production:latest eu.gcr.io/$GCLOUD_PROJECT/$CIRCLE_PROJECT_REPONAME:branch-$CIRCLE_BRANCH
+              docker tag production:latest eu.gcr.io/$GCLOUD_PROJECT/$CIRCLE_PROJECT_REPONAME:build-$CIRCLE_BUILD_NUM
+
```
rjw57: Was this related to #95? If so, I think this is avoid the issue rather than fixing it. I think the "correct" fix is removing the ``branch-...`` tag as outlined in that issue.

frontend/src/components/ProfileButton.js
rjw57: This needs removing.

smsjwplatform/jwplatform.py
```diff
@@ -101,6 +114,10 @@ def check_user_access(self, user):
                 return True
         raise ResourceACLPermissionDenied()
 
+    def get_poster_url(self, width=720):
+        return 'https://cdn.jwplayer.com/thumbs/{key}-{width}.jpg'.format(
```
rjw57: This should use the ``JWPLATFORM_API_BASE_URL`` setting.

PR #110, opened 2018-07-03
Add legacy sms redirect for media item page

### LINKED ISSUE #82: Create a legacy individual media item view endpoint
- This endpoint should translate old sms media ids into jwplayer ids
- It should redirect to the new individual media view if found
- it should redirect to the old sms if the media item is not found

rjw57:
Add an endpoint under /legacy for handling individual media pages. It maps the SMS media id to a JWPlayer key and redirects to the new media page. If the media item could not be found or if the ACL disallows viewing, the user is redirected back to the original SMS application to let it deal with the problem.

Closes #82

### DIFF
--- a/legacysms/tests/test_redirect.py
+++ b/legacysms/tests/test_redirect.py
@@ -62,3 +62,20 @@ class DownloadTests(TestCase):
         """Passing non-integer clip id raises an error."""
         with self.assertRaises(ValueError):
             redirect.media_download(1234, 'some/malicious/path', 'mp4')
+
+
+@override_settings(
+    LEGACY_SMS_REDIRECT_FORMAT='{url.scheme}://test.{url.netloc}{url.path}',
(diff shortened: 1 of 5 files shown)


legacysms/tests/test_redirect.py
```diff
@@ -62,3 +62,20 @@ def test_requires_integer_clip_id(self):
         """Passing non-integer clip id raises an error."""
         with self.assertRaises(ValueError):
             redirect.media_download(1234, 'some/malicious/path', 'mp4')
+
+
+@override_settings(
+    LEGACY_SMS_REDIRECT_FORMAT='{url.scheme}://test.{url.netloc}{url.path}',
```
abrahammartin: nice!

msb:
deployed/tested on https://development.media.cam.ac.uk/

tested for success, 404, & 403.

PR #136, opened 2018-07-10
Add mediaplatform and jwplatform application and legacysms models referencing said

rjw57:
This PR adds a new ``mediaplatform`` application and associated models (#120), a ``jwlplatform`` application and associated models (#129) and models referencing the ``mediaplatform`` models to the ``legacysms`` application (#121). It also expands the ACL scheme from (#127) into a new "permission" model but does not actually provide any of the logic specified in #127.

No indexes have been added to the models yet. We'll add them as we need them.

See the individual commit messages for more detail.

Closes #120. Closes #121. Closes #129.

Issue #127, opened 2018-07-04
Add ACL support for mediaplatform


rjw57:
From #128

> The permission model described here has been superseded by that in #138.

Make use of ``mediaplatform.MediaItem.acl`` field. The object is a JSON object with the following shape and all properties are optional.
```js
{
    isPublic: true,
    isSignedIn: true,
    institutionIds: ['UIS', 'ENG', /* etc */],
    lookupGroups: [1234, 5678, /* etc */],
    crsIds: ['spqr1', 'abc2', /* etc */],
}
```
The database ``WHERE`` clause should therefore return an item if:
* ``isPublic`` is ``true``, OR
* ``isSignedIn`` is ``true`` AND the user is not anonymous, OR
* the intersection of the user's institutions and ``institutionIds`` is non-empty, OR
* the intersection of the user's lookup groups and ``lookupGroups`` is non-empty, OR
* the user's CRSid appears in ``crsIds``.

Add a helper to the model's object manager to add an ``allowed_by_acl`` annotation to a query set which takes the current user's crsid, lookup groups and institutions. The ``allowed_by_acl`` field should be TRUE iff the user is allowed access by the ACLs.

Add a helper to the model's object manager which takes a user and returns a queryset filtered so that only those objects visible to that user are returned.

rjw57:
The ACL description above has been superseded by #138 which has a more traditional model-based permission.

PR #137, opened 2018-07-11
Add mediaplatform, mediaplatform_jwp applicationand legacysms models referencing said

### LINKED ISSUE #120: Add mediaplatform application
From #128:

Add a new Django application, ``mediaplatform`` with the following models:
```python
class MediaItem(Model):
    id = UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    title = TextField()
    description = TextField()
    acl = JSONField()
    # ... metadata we care about ...
    created_at = DateTimeField(auto_now_add=True)
    updated_at = DateTimeField(auto_now=True)
    deleted_at = DateTimeField()

class Collection(Model):
    id = UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
    title = TextField()
    description = TextField()
    media_items = ArrayField(ForeignKey(MediaItem))
    created_at = DateTimeField(auto_now_add=True)
    updated_at = DateTimeField(auto_now=True)
    deleted_at = DateTimeField()
```
(As time goes on we may wish to update/augment these models but this is the minimal set of information for the moment.)

### LINKED ISSUE #121: Add models to legacysms application
From #128:

Add models to represent SMS media items and collection to the ``legacysms`` application. They should include all the columns we wish to preserve from the original SMS.

The models should have the following form:
```python
class MediaItem(Model):
    id = BigIntegerField(primary_key=True, editable=False)  # "media_id" from SMS
    item = OneToOneField('mediaplatform.MediaItem', related_name='sms_media_item')
    last_updated_at = DateTimeField()

class Collection(Model):
    id = BigIntegerField(primary_key=True, editable=False)  # "collection_id" from SMS
    collection = OneToOneField('mediaplatform.Collection', related_name='sms_collection')
    last_updated_at = DateTimeField()
```

### LINKED ISSUE #129: Add jwplatform application
From #128:

Add a new application, ``jwplatform``. Ultimately, a lot of ``smsjwplatform`` will be copied into it but we need to start a new one so that the model tables have sensible names.
```python
class Video(Model):
    key = CharField(primary_key=True)
    item = OneToOneField('mediaplatform.MediaItem', related_name='jwp_video')
    updated = IntegerField()  # timestamp from JWP
```

rjw57:
> Note: this was re-opened from #136 after re-naming the ``jwplatform`` application to ``mediaplatform_jwp``.

This PR adds a new ``mediaplatform`` application and associated models (#120), a ``mediaplatform_jwp`` application and associated models (#129) and models referencing the ``mediaplatform`` models to the ``legacysms`` application (#121). It also expands the ACL scheme from (#127) into a new "permission" model but does not actually provide any of the logic specified in #127.

No indexes have been added to the models yet. We'll add them as we need them.

See the individual commit messages for more detail.

Closes #120. Closes #121. Closes #129.


mediaplatform/models.py
```diff
@@ -0,0 +1,146 @@
...
+    is_signed_in = models.BooleanField(default=False)
+
+
+class MediaItem(models.Model):
+    """
+    An individual media item in the media platform.
+
+    """
+    #: Primary key
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
abrahammartin: This should be shorter to reference it in the url or are you thinking in doing it at runtime?

mediaplatform/models.py
```diff
@@ -0,0 +1,146 @@
...
+    deleted_at = models.DateTimeField(null=True, blank=True)
+
+
+class Collection(models.Model):
+    """
+    A collection of media items.
+
+    """
+    #: Primary key
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
abrahammartin: same comment as with media

mediaplatform/models.py
```diff
@@ -0,0 +1,146 @@
...
+    #: Collection description
+    description = models.TextField(help_text='Description for collection')
+
+    #: :py:class:`~.MediaItem` objects which make up this collection. Postgres does not (currently)
+    #: allow array elements to have a foreign key constraint added to them so we need to represent
+    #: the links as bare UUIDs. The upshot of this is that code which uses this field needs to
+    #: handle the (rare) case that a UUID in the list does not correspond to a current video.
+    #: YouTube, as an example, has this problem as well since videos in playlists are sometimes
+    #: replaced by a "deleted video" placeholder.
+    media_items = pgfields.ArrayField(
```
abrahammartin: why not just a foreingkey in media_item? Is it because we may want to preserve the order? I think is worth adding it to the comments.

legacysms/models.py
```diff
@@ -1,3 +1,29 @@
...
 
-# Create your models here.
+
+class MediaItem(models.Model):
+    #: SMS media id
+    id = models.BigIntegerField(primary_key=True, editable=False, help_text='Legacy SMS media id')
+
+    #: Corresponding :py:class:`mediaplatform.MediaItem`. Accessible from the
+    #: :py:class:`mediaplatform.MediaItem` model via the ``sms`` field.
+    item = models.OneToOneField('mediaplatform.MediaItem', related_name='sms',
```
abrahammartin: currently we can't delete videos from the SMS via this method, models.CASCADE will make it disappear form the website but not from the SMS. It's a valid option but we need to think if that then will break our sync and it will try it to upload it again

legacysms/models.py
```diff
@@ -1,3 +1,29 @@
...
+    #: The last updated at time from the legacy SMS. Used to determine which items need updating.
+    last_updated_at = models.DateTimeField(help_text='Last updated at time from SMS')
+
+
+class Collection(models.Model):
+    #: SMS collection id
+    id = models.BigIntegerField(primary_key=True, editable=False,
+                                help_text='Legacy SMS collection id')
+
+    #: Corresponding :py:class:`mediaplatform.Collection`. Accessible from the
```
abrahammartin: same comment as with media

mediaplatform_jwp/models.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    A JWPlatform video resource.
+
+    """
+    #: JWPlatform video key
+    key = models.CharField(primary_key=True, max_length=64, editable=False)
+
+    #: Corresponding :py:class:`mediaplatform.MediaItem`. Accessible from the
+    #: :py:class:`mediaplatform.MediaItem` model via the ``jwp`` field.
+    item = models.OneToOneField('mediaplatform.MediaItem', related_name='jwp',
```
abrahammartin: same comment as with sms legacy

mediaplatform_jwp/models.py
```diff
@@ -0,0 +1,18 @@
...
+                                on_delete=models.CASCADE)
+
+    #: The updated timestamp from JWPlatform. Used to determine which items need updating.
+    updated = models.BigIntegerField(help_text='Last updated timestamp')
```
abrahammartin: why a timestamp instead of a datetimefield?

mediaplatform/models.py
rjw57: I think the URL encoding is a separate problem. Base64 encoding the UUIDs should lead to URLs which look like https://alpha.media.cam.ac.uk/media/IeGheechaegool9ahth7ee which I think is probably OK. We could use a random 64-bit integer instead which would lead to URLs like https://alpha.media.cam.ac.uk/media/ool2ieFe0e. I'm neutral on which we choose so I'm happy to do whatever you'd prefer.

mediaplatform/models.py
rjw57: See reply above.

mediaplatform/models.py
rjw57: As a separate issue, I think it'd be worth running the Base64 encoded ids through, e.g., https://github.com/Volnus/YouTube-Blacklist to avoid accidentally having media with URLs containing bad words.

mediaplatform/models.py
rjw57: 1) If a foreign key in media_item then a media item can be in one and only one collection which seems a needless restriction.
2) If a foreign key in media_item then there is no ordering

Looking at the way the SMS currently does it, there is a separate table mapping media to collections with an index field. Although this would work, it seems ugly. If you're unhappy with losing the foreign key constraint, I'd use a many-to-many mapping table like this instead.

legacysms/models.py
rjw57: At no point will the UMP become the source of truth for the SMS so at no point will deleting a video from UMP be expected to delete it from the SMS. Rather, when the SMS ceases to be the source of truth, a video deleted on the UMP will remain on the SMS because they are unconnected systems. While the SMS is still the source of truth for the UMP, "deleting a video" should be done on the SMS and then synchronised to the UMP.

As I understand it, therefore, this is working as designed.

legacysms/models.py
rjw57: Same reply :).

mediaplatform_jwp/models.py
rjw57: A similar reply to the one about SMS but with the subtlety that it is unclear to me if it is ever valid to have a video on JWPlatform which has no corresponding ``mediaplatform.MediaItem`` entry. I've assumed not but my assumption could be mistaken.

mediaplatform_jwp/models.py
rjw57: Because that's what JWP uses and we want to compare apples and apples when determining what's updated.

mediaplatform/models.py
rjw57: > I'm neutral on which we choose so I'm happy to do whatever you'd prefer.

Actually, I'm coming down on the side of a random 64-bit integer as providing a nicer compromise so I think I'll re-work this to use random 64-bit integer ids. I'll work the Birthday Paradox probabilities first though.

Alternatively, we could use an approach like https://wiki.postgresql.org/wiki/Pseudo_encrypt or https://wiki.postgresql.org/wiki/Pseudo_encrypt_constrained_to_an_arbitrary_range but I think that's becoming needlessly complex.

rjw57: So, according to the table at https://en.wikipedia.org/wiki/Birthday_problem#Probability_table, to get a 1 in a million chance of collision with 64-bit ids we'd need around 6 million media items/collections. For 128-bit UUIDs, that becomes 2.6 x 10^16.

Further, collision only actually matters if we use 64-bit ids for more than one table (e.g. media items and collections). Within a table, Postgres saves us from collision. With that the probability that a) a collision occurs and b) it happens to be between a collection and a media item probably becomes tolerably small.

rjw57:
@abrahammartin OK, I've updated the PR now after your comments. Most notably, I've updated the primary key of the various ``mediaplatform`` models to be an opaque token which is generated via the builtin ``secrets`` module. This should simplify URL encoding since the primary key can be interpolated directly into the URL.

I've also changed the ``on_delete`` behaviour for SMS and JWP models to "SET_NULL". This essentially makes the database neutral on whether there ever can be a video on JWP or the SMS which we shouldn't know about.

The full diff from the original PR is visible here:

https://github.com/rjw57/sms-webapp/compare/issue-121-add-models-v1...rjw57:issue-121-add-models?expand=1

PR #138, opened 2018-07-11
Fixes for mediaplatform models

rjw57:
Since there were no tests, we did not discover that the models could not actually be created since the primary key field was too small :S. Address this issue and fix up some others found along the way. The remaining commit increases how lenient we are in what we accept in the database. The reason for this initial leniency is that our models will be being populated from JWP which itself was populated from the SMS. We cannot guarantee that all fields will always have values so, for the moment, err on the side of caution.

Since migrations were starting to make the coverage reports noisy, remove them from coverage calculations.


abrahammartin:
Deployed to https://development.media.cam.ac.uk/ but execute migration fails

rjw57:
> Deployed to https://development.media.cam.ac.uk/ but execute migration init container fails

I deployed a change including #138 to development earlier so suspect that this might be as simple as rolling back the (later) migration and re-deploying.

That being said, this change is deployed (by way of #142) to development already, so you can try it out there.

Issue #26, opened 2018-01-08
Secrets in settings


abrahammartin:
https://github.com/uisautomation/sms-webapp/blob/ab84baec44c2a45327c23658ee77bd5ac257212f/smswebapp/settings/base.py#L11

Remove all secrets from settings, load the from env vars, update docker-compose to add them to dev env, and add checks.

rjw57:
I'll prepare a branch based on #24 since that did quite the re-jig in this area.

rjw57:
Implemented on this branch: https://github.com/rjw57/sms-webapp/tree/remove-hardcoded-secret-key. I'll hold off on opening a PR until #24 is merged so that I can re-base on any changes in #24 which arise.

PR #139, opened 2018-07-16
compose: fix lookupproxy deployment

rjw57:
The lookupproxy compose configuration was only semi-functional. Since we're going to need to re-deploy the lookupproxy in development *anyway* to bring in the django-automationoauth fixes, fix up the configuration so that it can use the upstream production image.

This should further reduce CircleCI build times and also means that lookupproxy actually works in development. Previously it didn't since the OAuth2 configuration was not being set.


abrahammartin:
This is relying on the docker hub's lookupproxy, the last time we checked that wasn't working properly and it was not building for every push to master and we had to quick it manually. Has this changed now?

rjw57:
It seems to have worked last time I checked, but in any case if the lookupproxy image isn't building, then that's a separate issue: https://github.com/uisautomation/lookupproxy/issues/27

abrahammartin:
We can test with https://github.com/uisautomation/lookupproxy/pull/26

rjw57:
True, I just added a new automated build rule for a new branch and was going to push that branch to see, but merging uisautomation/lookupproxy#26 would work just as well. I just worry about merging that because we've got no way to safely test it at the moment.

PR #142, opened 2018-07-16
Update jwpfetch command to populate database objects from the cached resources

### LINKED ISSUE #130: jwpfetch command creates media items and legacysms media items
From #128 

Update the ``jwpfetch`` management command to:

* Create/update ``mediaplatform.MediaItem`` objects (using the ACL scheme in #127)
* Create/update ``legacysms.MediaItem`` objects
* Create/update ``jwplatform.Video`` objects

rjw57:
Add a new module, mediaplatform_jwp.sync, which contains a function which takes the current set of cached JWP resources and, atomically, will reconcile changes with the database objects.

The function is complex but is heavily commented so it can be verified bit by bit. Similarly, there are quite a few tests for the synchronisation code to try to make sure that all the code paths are tested, at least.

The jwpfetch management command is updated to synchronise the database after fetching all the JWP resources. It's given two new flags to control whether the JWP resources are actually fetched and whether all objects in the database are updated irrespective of update timestamp. See the jwpfetch documentation for details on when this can be useful.

Closes #130


rjw57:
Now rebased so that the PR doesn't include the commits from #130.

mediaplatform_jwp/sync.py
```diff
@@ -0,0 +1,330 @@
...
+
+    # Add the corresponding media item link to the JWP videos.
+    for key, item in jwp_keys_and_items:
+        jwpmodels.Video.objects.filter(key=key).update(item=item)
+
+    # 4) Update metadata for changed videos
+    #
+    # After this stage, all mediaplatform.MediaItem objects whose updated_at field precedes their
+    # corresponding JWP video's updated timestamp + MEDIA_ITEM_FRESHNESS_THRESHOLD will have their
+    # metadata updated from the JWP video's custom props.
```
msb: can we also note here that legacymodels.MediaItem are synchronised as well

msb:
sync() appears not to respect changes to updated flag

PR #154, opened 2018-07-20
Convert UI to use MediaItem objects

### LINKED ISSUE #126: Update ui to use mediaplatform.MediaItem objects
From #128

The UI should stop using ``smsjwplatform.CachedResource`` objects and instead use ``mediaplatform.MediaItem`` objects. It can then use the ``jwp_video`` related object.

Remove ``smsjwplatform.CachedResource`` model and stop ``jwpfetch`` updating them.

rjw57:
> PR #142 should be merged before this one is as this PR has the changes from #142 merged.

This PR replaces the media item views and underlying API views to use MediaItem as the source of truth rather than CachedResource. The actual amount of code change is quite small but these commits are rather large because the test suite needed to be fixed up to create MediaItem fixtures as opposed to CachedResource fixtures.

Commit 19b0202 does most of the work on the database model which is required by adding a custom manager and custom query set implementation which makes it hard to accidentally include deleted videos if you don't intend to and to provide a central location to add the database ACL queries. This commit intentionally does not add any ACL checking into this function since that is part of a separate story this sprint. Instead, various TODOs are left where ACL support should be added.

Commit 9e6d3fa a small frontend bug.

Commit c35be11 updates the UI and API views to use MediaItem objects when rendering their response. Moving to using database models means we can make use of DRF's ModelSerializer and DRF's built in ordering and pagination classes. We implement a custom search filter to allow searching by tag/keyword.

Commit d5b6118 experiments with changing the schema of the JSON returned form the API to match the JSON-LD schema for a "Video" object. (Taken from Google's documentation for said.) Again, there's not much code in that commit but a fair amount of churn because all the tests need fixing up.

Closes #126

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -24,11 +24,11 @@ class MediaSerializer(serializers.Serializer):
     An individual media item.
 
     """
-    id = serializers.CharField(source='key', help_text='Unique id for the media')
+    id = serializers.CharField(help_text='Unique id for the media')
     title = serializers.CharField(help_text='Title of media')
     description = serializers.CharField(help_text='Description of media')
-    published_at_timestamp = serializers.IntegerField(
-        source='date', help_text='Unix-style UTC timestamp of publication time')
+    published_at = serializers.DateTimeField(
+        help_text='Publication time')
     poster_image_url = serializers.SerializerMethodField(
         help_text='A URL of a thumbnail/poster image for the media'
     )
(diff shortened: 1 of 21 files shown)


rjw57:
@abrahammartin conflicts resolved but I'd wait until #142 is merged and then I'll rebase to make the PR cleaner.

PR #157, opened 2018-07-20
Respect view permissions in UI and API

### LINKED ISSUE #119: Fix mobile view of individual media items
- Wrong margins
- Download media and statistic links should be buttons so that in the mobile view they can be seen in proper format and are easy to click.

### LINKED ISSUE #127: Add ACL support for mediaplatform

rjw57:
This PR implements the view permissions (ACLs) on media items imported into the database. On the advice of the lookup maintainer, it also changes lookup groups to be stored as strings rather than numeric fields. (See the message from 0fb78d5 for details.)

The bulk of the work is in c0577ec, d4eaa62 and 390c1a4 which implements the viewable_by_user() and editable_by_user() filter for MediaItems. This commit seems large because of all the test churn but the actual code is small.

Along the way, it was opportune to address #119 as well in commit 5067350.

> **UPDATE:** I've updated the PR to include edit permissions and some more tweaks to the media platform model which were exposed when writing the upload UI code.

Closes #127. Closes #119.


abrahammartin:
more conflicts :(

rjw57:
> more conflicts :(

Fixed. I ran ``git merge -Xours master`` which should Do The Right Thing

rjw57:
> All good, but in order to close #119 it needs to address the buttons as well.

Should be done in d5c04fd.

PR #159, opened 2018-07-25
Re-work API/UI interaction

rjw57:
This PR tidies up our API/UI split a little bit and introduces the ability to create/update media items via the API.

Since this PR changes the API view classes to be a ListCreate and RetrieveUpdate view, we can no-longer simply derive from them for the UI. We address this by splitting the specialisations for Media item list and detail views into their own mixins and having the UI and API views inherit from those mixins.

### DIFF
--- a/api/permissions.py
+++ b/api/permissions.py
@@ -1,0 +1,56 @@
+from rest_framework import permissions
+
+import mediaplatform.models as mpmodels
+
+
+class MediaItemPermission(permissions.BasePermission):
+    """
+    A permission which allows a user access to "safe" HTTP methods if they have the view permission
+    and *additionally* requires the edit permission for "unsafe" HTTP methods.
+
+    The "additional" above implies that a user cannot modify a video if they cannot view it even if
+    they have the edit permission.
+
+    Come what may, a user can do *nothing* to a deleted video.
+
+    """
+    def has_permission(self, request, view):
+        if request.method in permissions.SAFE_METHODS:
+            return True
+
+        # Only signed in users have permissions to perform "unsafe" operations
+        if request.user is None or request.user.is_anonymous:
+            return False
+
+        return True
+
+    def has_object_permission(self, request, view, obj):
+        return self.has_media_item_permission(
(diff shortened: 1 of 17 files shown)


api/serializers.py
```diff
@@ -72,6 +72,33 @@ class Meta:
...
+            if not request.user.is_anonymous:
+                # Due to Django ORM oddness, we need to re-fetch the object to correctly modify
+                # permissions otherwise the ORM gets confused
+                new_item = (
+                    mpmodels.MediaItem.objects.all()
+                    .only()
+                    .select_related('view_permission', 'edit_permission')
+                    .get(id=new_item.id)
+                )
+                new_item.view_permission.crsids.append(request.user.username)
```
abrahammartin: doesn't the post_save does this at this moment? Or is this to have more completeness?

api/permissions.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def has_media_item_permission(self, request, obj, requires_view, requires_edit):
+        # Sanity check that the passed object is a MediaItem.
+        assert isinstance(obj, mpmodels.MediaItem)
+
+        # Ask the DB what the user can do with this item.
+        item = (
+            mpmodels.MediaItem.objects.all()
+            .only()  # Don't fetch all fields; we only use the viewable and editable annotations
+            .annotate_viewable(request.user)
```
abrahammartin: is it worth checking if the item already comes annotated and in this case then no need to re-fetch it from the database? MediaItem and MediaItemList mixins seem to annotate these elements. Although the saving doesn't seem much (1 query per item), would it amount to something significant for lists?

rjw57: It is worth doing this. Wilco.

api/serializers.py
rjw57: The post_save hook merely makes sure that an associated permission object is created in the database. This makes sure it has the current user set on the permission. (Note that no permission object is created here.) Since this may be a common pattern, I'll hoist it into an object manager method.

Issue #150, opened 2018-07-17
Display video statistics


rjw57:
Provide some form of statistics view for a media item under http://hostname/media/[id]/analytics. (Use a modern naming convention in the URL at least.) Use a solution similar to that in https://tlscerts.uis.cam.ac.uk/stats to display video views by date.

This should, initially, use the stats from the ``stats.media_stats_by_day`` table in our database.

PR #165, opened 2018-07-26
Media item analytics

msb:
Implements a video statistics page using legacy statistics and a Google Charts Annotation Chart. While we don't use the annotations feature, it's useful because of it's Pan/Zoom function.

closes: #150


frontend/src/index.css
```diff
@@ -1,5 +1,20 @@
...
+  display: none;
 }
+/* Correct the legend & zoomControls font */
+.google-visualization-atl .legend,
+.google-visualization-atl .zoomControls {
+  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
+}
+
+/* Missing bottom border */
+#reactgooglegraph-1_AnnotationChart_rangeControlContainer {
```
rjw57: Why not ``.rangeControl``?

frontend/src/index.css
```diff
@@ -1,5 +1,20 @@
...
+/* Hide the useless displayZoomButtons */
+#reactgooglegraph-1_AnnotationChart_zoomControlContainer_1-hour,
+#reactgooglegraph-1_AnnotationChart_zoomControlContainer_1-day,
+#reactgooglegraph-1_AnnotationChart_zoomControlContainer_5-days {
+  display: none;
 }
+/* Correct the legend & zoomControls font */
+.google-visualization-atl .legend,
+.google-visualization-atl .zoomControls {
+  font-family: "Roboto", "Helvetica", "Arial", sans-serif;
```
rjw57: I think that if we wrap the chart in a ``Typography`` component, this should be fixed. E.g.:
```patch
diff --git a/frontend/src/pages/AnalyticsPage.js b/frontend/src/pages/AnalyticsPage.js
index e457286..14e3ca2 100644
--- a/frontend/src/pages/AnalyticsPage.js
+++ b/frontend/src/pages/AnalyticsPage.js
@@ -47,11 +47,11 @@ class AnalyticsPage extends Component {
                 Viewing history (views per day)
               </Typography>
             <Grid item xs={12}>{
-              chartData.length > 1 ? <Chart
+              chartData.length > 1 ? <Typography variant='body1' component='div'><Chart
                 chartType="AnnotationChart"
                 data={chartData}
                 options={{fill: 100, colors: ['#EF2E31']}}
-              /> : <Typography variant="subheading">
+              /></Typography> : <Typography variant="subheading">
                 There is no data available for the media
               </Typography>
             }</Grid>
```

frontend/src/pages/MediaPage.js
```diff
@@ -35,7 +34,7 @@ const MediaPage = ({ mediaItem, classes }) => (
         </Grid>
         <Grid item xs={6} style={{textAlign: 'right'}}>
           <Typography variant="subheading">
-            <a className={ classes.link } href={mediaItem.statsUrl}>
+            <a className={ classes.link } href='./analytics'>
```
rjw57: Can we not just set ``mediaItem.statsUrl`` using Djano's ``reverse()`` helper?

ui/views.py
```diff
@@ -24,3 +26,48 @@ def get(self, request, media_key):
...
+    def get(self, request, media_key):
+
+        response = super().get(request, media_key)
+
+        min_date = datetime.datetime.max
+        max_date = datetime.datetime.min
+        summed_by_date = {}
+
+        # Here we sum up all views for a particular day (irrespective of other variable) and
+        # calculate the min and max dates.
```
rjw57: I feel this is exactly the sort of thing the database should be doing...

msb: as discussed - will move this into client

legacysms/routers.py
```diff
@@ -0,0 +1,31 @@
...
+
+        return 'default'
+
+    def db_for_write(self, model, **hints):
+        """
+        The only stats schema table defined is MediaStatsByDay,
+        so route queries on this to the stats DB.
+        """
+        if legacysms.models.MediaStatsByDay == model:
+            return 'stats'
```
rjw57: Since ``db_for_{read,write}`` are identical, could we replace them both with a call to a single private method?

ui/views.py
```diff
@@ -19,8 +23,56 @@ class MediaView(api.views.MediaView):
...
+
+        day_count = (max_date - min_date).days + 1
+
+        analytics = [
+            {"date": date.timestamp(), "views": summed_by_date.get(date, 0)}
+            # Note we also add a zero data-point at either end of the data which makes the graph
+            # look better in case of a single data-point.
+            for date in (min_date + datetime.timedelta(n) for n in range(-1, day_count + 1))
+        ]
+"""
```
rjw57: If this is not useful, can we just delete it?

frontend/src/api.ts
```diff
@@ -208,7 +208,7 @@ export const mediaCreate = (body: IMediaCreateResource) : Promise<IMediaResource
 /** Retrieve a media resource. */
 export const mediaGet = (id: string) : Promise<IMediaListResponse | IError> => {
   const resource = resourceFromPageById(id);
-  if(resource) { return Promise.resolve(resource); }
+  if (resource) { return Promise.resolve(resource); }
```
rjw57: No space here, please. We try hard to pretend JS is C :).

frontend/src/pages/MediaPage.js
```diff
@@ -94,10 +93,10 @@ const MediaPageContents = ({ item, classes }) => {
           </Grid>
           <Grid item xs={12} sm={6} md={3} lg={2} style={{textAlign: 'right'}}>
             {
-              item && item.links && item.links.legacyStatisticsUrl
+              item
```
rjw57: ``item && item.id``?

ui/views.py
```diff
@@ -22,3 +25,18 @@ class MediaView(apiviews.MediaItemMixin, generics.RetrieveAPIView):
...
+    Extends the DRF's media item analytics view.
+
+    """
+    renderer_classes = [TemplateHTMLRenderer]
+
+    template_name = 'ui/analytics.html'
+
+    def get(self, request, pk):
+        response = super().get(request, pk)
+        return Response({'analytics_json': json.dumps(response.data)})
```
rjw57: Can we use the ``to_json`` filter instead? Otherwise the analytics will have the same XSS fail if there is a user-controllable string in them in the future.

frontend/src/pages/AnalyticsPage.js
```diff
@@ -0,0 +1,197 @@
...
+          </Grid>
+          <Grid container justify='space-between' spacing={16}>
+            <Grid item xs={12} sm={6} md={3} lg={2}>
+              <Button
+                component='a' variant='outlined' className={ classes.link }
+                href={'/media/' + pk} fullWidth
+              >
+                Media Item
+                <ArrowUpwardIcon className={ classes.rightIcon } />
+              </Button>
```
rjw57: Remove this button.

frontend/src/pages/AnalyticsPage.js
```diff
@@ -0,0 +1,197 @@
...
+            </Grid>
+            <Grid item xs={12} sm={6} md={3} lg={2} style={{textAlign: 'right'}}>
+              <Button component='a' variant='outlined' className={ classes.link } fullWidth
+                href={mediaItem && mediaItem.links && mediaItem.links.legacyStatisticsUrl}
+              >
+                Legacy Statistics
```
rjw57: Change this label to "Statistics on SMS Site"

PR #185, opened 2018-08-01
Add Channel model

### LINKED ISSUE #177: Add the concept of a channel
This comes from issue #169 

We need to create a new Channel model that will work like the Youtube model. MediaItems will start to belong to a Channel. And deleting the Channel will cause the MediaItem to be deleted or transfer ownership to another Channel.

Take important fields from Collections that we want to preserve in Channels.

No Channel alias by now

rjw57:
This PR adds a new Channel model as specified in #177. Not much functionality is present and, importantly, we do not try to move the edit permission from media items to channels in this PR. To deal with migration, we create a special "orphan" channel which takes all the media items not in any other channel. Initially all the media items become part of this channel but in a future PR we will teach ``jwpfetch`` to assign media items to the correct channels.

Closes #177


msb:
Testing: Channel.description is missing in the admin panel

PR #187, opened 2018-08-02
fix JSON encoding XSS vunerability

rjw57:
The to_json filter tag did not escape HTML special characters within strings meaning that any resource which had, e.g., a description with closing <script> tag in it could trivially perform a XSS attack.

Update the tag with an implementation inspired by the new json_script tag from Django 2.1. We cannot use the json_script tag directly because we add more information to our script tags than simply the id.

### DIFF
--- a/ui/templates/ui/media.html
+++ b/ui/templates/ui/media.html
@@ -3,9 +3,9 @@
 {% block title %}{{ resource.title }}{% endblock %}
 {% block extrabody %}
 <script type="application/ld+json">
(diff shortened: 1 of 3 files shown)


abrahammartin:
2.1 is now in master, so you are free to use 2.1 new features

rjw57:
Unfortunately json_script isn't a drop-in replacement. It's worth seeing if we can migrate to it in future but, at least in the case of JSON-LD, we still need a customised version since json_script doesn't allow us to set the type of the script tag.

PR #188, opened 2018-08-02
port to Django 2.1 and Python 3.7

rjw57:
Update the portions of our development and production packaging which specify Django 2.0 and Python 3.6 to talk about Django 2.1 and Python 3.7. Bump the minimum required version of django-ucamwebauth to be Django 2.1 compatible.

Make tox neutral on the precise version of Python it uses (beyond Python3) since we always run tox inside a container and it should just use whatever Python 3 is given to it.


abrahammartin:
Deployed to development environment. Tested, and all working as expected. Merging.

PR #189, opened 2018-08-05
frontend: tidy up display on small screens

rjw57:
This commit collects some work together to improve the display on small screens. Firstly, the tight margin of the AppBar is tidies up so that the "sign in" button lines up visually with the right margin. The search bar now disappears on small screens with only the search icon remaining.  When clicked, the search icon transitions the bar to a special "search" state where the search form is shown.

Replace the shield image with "Media Platform" since it better defines the site.

This is up for debate. I'm happy to keep the shield but, try as I might, I cannot make the words "Media Platform" look good next to it. We may want to investigate having the branding elsewhere.


abrahammartin:
requested to Joe if he can design us a logo

abrahammartin:
Pending from @joeadams04 to send us the new logo for the Media Platform.

msb:
tested on https://test.media.cam.ac.uk/

PR #194, opened 2018-08-06
Synchronise JWP channels to mediaplatform Channel objects

### LINKED ISSUE #178: jwpfetch move jwplayer playlists to Channels

rjw57:
This PR updates jwpfetch to synchronise JWP channels to mediaplatform channels using the list of sms media ids in the channel custom prop as the source of truth.

When jwp is run, channels are created for each JWP channel and the media items are populated using the sms_media_id custom prop. Basic infomration such as title and description are also copied.

This PR re-uses much of the code from the media item synchronisation but genericises it a little. In particular we still use the Cachedresource model for the synchronisation since the associated code is tested well in production.

Note that the ACL for the channel is not copied over at the moment. SMS collections are being split into "channels" and "playlists". A "view" permission makes more sense for a playlist than a channel; the existence of a channel should not be a secret but individual playlists and media items within it may be.

This breaks down if the ACL for a SMS collection should be considered in addition to the individual media item. Rather than making the (possibly wrong) decision about how SMS ACLs are mapped into our new world, this PR ignores the problem and waits until we also have a Playlist model.

Closes #178


rjw57:
Closing to see if codecov will comment on a PR from the original repo.

PR #195, opened 2018-08-06
Synchronise JWP channels to mediaplatform Channel objects

### LINKED ISSUE #178: jwpfetch move jwplayer playlists to Channels

rjw57:
This PR updates jwpfetch to synchronise JWP channels to mediaplatform channels using the list of sms media ids in the channel custom prop as the source of truth.

When jwp is run, channels are created for each JWP channel and the media items are populated using the sms_media_id custom prop. Basic infomration such as title and description are also copied.

This PR re-uses much of the code from the media item synchronisation but genericises it a little. In particular we still use the Cachedresource model for the synchronisation since the associated code is tested well in production.

Note that the ACL for the channel is not copied over at the moment. SMS collections are being split into "channels" and "playlists". A "view" permission makes more sense for a playlist than a channel; the existence of a channel should not be a secret but individual playlists and media items within it may be.

This breaks down if the ACL for a SMS collection should be considered in addition to the individual media item. Rather than making the (possibly wrong) decision about how SMS ACLs are mapped into our new world, this PR ignores the problem and waits until we also have a Playlist model.

Closes #178


abrahammartin:
> Note that the ACL for the channel is not copied over at the moment.

A Channel does not need a view permission and this will be up to each individual media item for the channel to show up or not (I think we agree on that). 

What I think this story needs to cover is the import of editable permissions from SMS Collection to MP Channels and ownership of the Channel/Collection (institution the channel/collection belongs to). 

Ownership can be extracted from the SMS Collection's institution and the editable permissions can also can be extracted from the SMS Collection field "editable group" which will usually be a lookup group. See for example: https://upload.sms.cam.ac.uk/collection/1084806

rjw57:
The editor group is already copied over in this PR. I'm not sure if institutional ownership has an equivalent in the data model we currently have.

rjw57:
I think this PR correctly moves over the permissions we're going to implement but I may have the wrong idea of what those permissions are. I've added a draft document at https://gist.github.com/rjw57/5a6f0fc3e03b7c1313b5af7e28515aa9 which outlines my understanding of the access control model.

mediaplatform_jwp/sync.py
```diff
@@ -314,6 +303,194 @@ def update_related_models_from_cache(update_all=False):
...
+        ]
+        channel.items.set(
+            mpmodels.MediaItem.objects.filter(sms__id__in=collection_media_ids).only('id')
+        )
+
+        # Update associated SMS collection (if any)
+        sms_collection_id = channel_data.collection_id
+        if sms_collection_id is not None:
+            # Get or create associated SMS collection. Note that hasattr is recommended in the
+            # Django docs as a way to determine isf a related objects exists.
```
msb: s/b if

smsjwplatform/management/commands/jwpfetch.py
```diff
@@ -36,41 +37,89 @@ def add_arguments(self, parser):
         parser.add_argument(
             '--sync-all', action='store_true', dest='sync_all',
             help='Synchronise all objects from JWP cache, not only those which have been updated')
+        parser.add_argument(
```
msb: Could you update the comments accordingly?

msb:
Testing: changes to CachedResource(CHANNEL) objects aren't being passed thru to the Channel objects

Issue #179, opened 2018-07-30
Create a Playlist model (essentially current Collections model except for the management ACLs)


abrahammartin:
Should have an order
Should have viewing but not management ACLs
Should belong to a channel

PR #198, opened 2018-08-07
Create Playlist model.

msb:
Creates the Playlist model with associated migrations, tests, and admin support. Not directly connected with the issue, I have added support for PUDB (https://documen.tician.de/pudb/) - a debugger that's easy to use with docker-compose. Can be reviewed commitwise.

closes: #179


mediaplatform/tests/test_models.py
```diff
@@ -350,6 +351,7 @@ def setUp(self):
 
     def test_creation_no_fields(self):
         """A Channel object should *not* be creatable with no field values."""
+        # TODO this should fail?
```
rjw57: Not sure. I *think* ``clean()`` should fail, but I don't think Django stops you creating things if they are not disallowed by the DB.

compose/development.yml
```diff
@@ -62,6 +62,8 @@ services:
         target: /usr/src/app
         read_only: true
       - "frontend_build:/tmp/frontend_build"
+    stdin_open: true
+    tty: true
```
rjw57: Can we add a comment to this explaining its necessity?

mediaplatform/migrations/0010_auto_20180807_0948.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.1 on 2018-08-07 09:48
```
rjw57: Can we give this migration a more informative name?

mediaplatform/migrations/0008_auto_20180806_1436.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 2.1 on 2018-08-06 14:36
```
rjw57: Can we give this a more informative name?

PR #201, opened 2018-08-07
Add a channel view to the UI and associated API views

### LINKED ISSUE #149: Provide channel view
Provide a view of channel under http://hostname/channel/[id]. The view should be similar to the latest media items on the main index page but be limited to a channel. Respect the view_permission of the channel.

rjw57:
This PR adds a view to the UI for channels and list, create, retrieve and update support for channels to the API. It takes the opportunity to genericise some code in the API a little so that the actual new code required to support the channel api is fairly minimal.

This can be tested by searching for a channel in the UI and clicking on the card.

Closes #149


api/serializers.py
```diff
@@ -177,3 +179,68 @@ class MediaAnalyticsListSerializer(serializers.Serializer):
...
     results = MediaAnalyticsItemSerializer(many=True, source='*')
+
+
+class ChannelSerializer(serializers.HyperlinkedModelSerializer):
+    """
+    An individual channel.
+
+    """
+    class Meta:
+        model = mpmodels.MediaItem
```
abrahammartin: doesn't this needs to be mpmodels.Channel?

PR #203, opened 2018-08-08
bugfix: we get 404 if the JW video is not yet available

msb:
We need to query the JW api to generate the download sources. However if the video is still transcoding this will fail and was resulting in a 404. Better to set the sources to none and log a warning.

### DIFF
--- a/api/tests/test_views.py
+++ b/api/tests/test_views.py
@@ -163,6 +163,17 @@ class MediaItemViewTestCase(ViewTestCase):
             response = self.view(self.get_request, pk=obj.id)
             self.assertEqual(response.status_code, 200)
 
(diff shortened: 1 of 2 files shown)


api/tests/test_views.py
```diff
@@ -172,7 +172,7 @@ def test_success_if_jw_item_not_yet_available(self):
         response = self.view(self.get_request, pk=item.id)
         self.assertEqual(response.status_code, 200)
         # sources cannot be generated and so are set as None
-        self.assertIsNone(response.data['sources'])
+        self.assertFalse(response.data['links']['sources'])
```
rjw57: Could we ``assertEquals(..., [])`` instead?

PR #209, opened 2018-08-13
Bugfix for "Unhandled Rejection" error on 6060

msb:
This error arises because the Page component is trying to query the API, which isn't available in the context of 6060. A better fix for this might be to generalise the api.resourceFromPageById() method for all api calls would allow you to set up dummy data for the components in 6060.


frontend/src/components/ItemMetadataForm.md
```diff
@@ -39,5 +39,5 @@ If you want to show an error state, set the errors object:
 const item = { title: 'The Dead Parrot Sketch' };
 const errors = { title: ['The parrot is alive.', 'It is sleeping.'] };
 
-<ItemMetadataForm item={ item } errors={ errors } />
+<ItemMetadataForm item={ item } errors={ errors } onChange={ _ => {} } />
```
rjw57: I think the correct fix is to make onChange not required and either
1. fix up the call sites so ``event => onChange(...)`` becomes ``event => onChange && onChange(...)``, or
2. give onChange the default value of ``() => null``.

frontend/src/api.ts
```diff
@@ -182,7 +182,10 @@ export const apiFetch = (
       return response.json().then(body => Promise.reject({
         body,
         error: new Error('API request returned error response'),
-      }));
+      })).catch(error => {
+        // tslint:disable-next-line:no-console
+        console.error('error parsing response:', error);
+      })
```
rjw57: It may be that it makes more sense to move ``Page`` into ``src/containers`` than to try and make it a true component.

PR #226, opened 2018-08-16
Add navigation drawer

rjw57:
As noted in #213, it would be nice to have a Material Design drawer for the application. This is especially true when we come to add "My channels"-style links, etc.

This is the output from a bit of playing I did yesterday on adding a navigation bar. It looks quite nice so I'm opening this PR as a bit of a "Brucey-bonus". We augment the profile view from the API to include a display name and the user's image from lookup (if set).


frontend/src/components/PlaylistList.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
msb: the example displays as blank for me

frontend/src/components/ResourceList.js
```diff
@@ -0,0 +1,115 @@
+import React from 'react';
```
msb: remove all refs to channel is the comments

msb:
these is an unused "import Button" in Page.js

PR #232, opened 2018-08-17
add Django robots.txt and sitemap framework

### LINKED ISSUE #173: robots.txt
forbid indexing test and development deployments

rjw57:
Add django-robots and django.contrib.sitemaps applications and provide some configuration for the sitemaps.

The sitemap is configured at the moment to include the 200 most recently updated media items and the index page. As we add channel views, etc, we should add them to the sitemap to help Search Engines out.

The actual configuration of the robots.txt is via the Djano sites framework and is performed in the admin so, once deployed, there is a chore to go and configure all the sites.

The sites framework records the domain name of the site in the database.  The default post_migration hook creates an "example.com" entry which also needs modifying in the admin on all the deployed sites.

Closes #173

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -15,6 +15,7 @@ iso-639
 python-dateutil
 django-crispy-forms
 django-filter
+django-robots
 
 # PRE-RELEASE WHITENOISE VERSION
 # We need at least version 4 of whitenoise to make use of the index_file
--- a/smswebapp/settings/base.py
+++ b/smswebapp/settings/base.py
@@ -25,6 +25,8 @@ INSTALLED_APPS = [
     'django.contrib.messages',
     'whitenoise.runserver_nostatic',  # use whitenoise even in development
     'django.contrib.staticfiles',
+    'django.contrib.sites',
+    'django.contrib.sitemaps',
 
     'automationcommon',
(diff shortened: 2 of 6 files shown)


abrahammartin:
All seems good but I think we need to add a complementary PR in media-deploy and/or a env variable to reject any indexing unless it is the production environment. Otherwise we have the risk of index non-prod environments by mistake. Something like always forbid to index except if it is the production one. The complementary PR in media-deploy could also include the data migration for the database with the correct domain name? So we avoid deploying incorrectly in the future (knowledge gets lost and we may forget to change example.com in the future)

rjw57:
Having a site specific migration is slightly going against the grain of the way this is supposed to work but we can probably work something out.

rjw57:
So, having pondered this, we can probably do it with a YAML fixture in the deploy repo and have the migration container also run load data. I *think* load data does the right thing if the data is already there.

abrahammartin:
Yes, something like that I was thinking of or perhaps an env variable like SITEMAP that is only True in media-deploy if production and then add to the script that gets executed everytime we start a container (that currently only executes migrations) a check that the database contains the right content regarding sitemap (correct URL and index everything in production, no index anything otherwise)

PR #238, opened 2018-08-20
Add download and embed views to API

rjw57:
This PR adds an /api/media/{pk}/source and /api/media/{pk}/embed path to the API which allows retrieving individual media source streams and embedding media within an IFrame. It fixes up the frontend and other API serialisers to use these endpoints.

It does not touch the legacysms code since that is blocked on #231 being merged. If this PR gets merged before #231, then #231 should be updated.

The first few commits are some simple housekeeping tasks noted while creating this PR. The later commits implement the bulk of the work. Care is taken to make sure the Swagger schema continues to document the API by adding custom schema for the .../source and .../embed endpoints.


api/views.py
```diff
@@ -178,6 +181,35 @@ def get_queryset(self):
         return super().get_queryset().select_related('upload_endpoint')
 
 
+class MediaItemEmbedViewInspector(inspectors.ViewInspector):
```
msb: seem to be missing imports for inspectors/openapi - if this has been addressed in a subsequent commit then ignore

api/views.py
```diff
@@ -295,3 +327,21 @@ class PlaylistView(PlaylistMixin, generics.RetrieveUpdateAPIView):
 
     """
     serializer_class = serializers.PlaylistDetailSerializer
+
+
+def exception_handler(exc, context):
```
msb: maybe better to put this in it's own module

api/serializers.py
```diff
@@ -253,22 +253,13 @@ def update(self, instance, verified_data):
 
 class MediaItemLinksSerializer(serializers.Serializer):
     legacyStatisticsUrl = serializers.SerializerMethodField()
-    embedUrl = serializers.SerializerMethodField()
```
msb: what's the reasoning behind removing the links dictionary?

msb:
Tested on https://development.media.cam.ac.uk:

- new endpoints
- client using changed endpoints
- new legacy/embed behaviour

Issue #181, opened 2018-07-30
Create and delete a playlist and reorder its associated videos


abrahammartin:
1) Create an empty playlist
2) Delete a playlist (does not delete its videos)
3) Reorder its associated videos

abrahammartin:
done now

PR #241, opened 2018-08-21
Allows the user to create an empty playlist.

msb:
The user can create an empty playlist in any channel they can edit. 
Partially addresses #181.
Can be reviewed commit-wise.


ui/urls.py
```diff
@@ -26,8 +27,17 @@
     path('media/<pk>/edit', views.MediaEditView.as_view(), name='media_item_edit'),
     path('media/<pk>', views.MediaView.as_view(), name='media_item'),
     path('channels/<pk>', views.ChannelView.as_view(), name='channel'),
+    path(
+        'create_playlist',
+        login_required(TemplateView.as_view(template_name="ui/create_playlist.html")),
+        name='create_playlist'
```
rjw57: To fit with other views, this should be called "playlist_create". (One could also argue that the upload view should be called "media_item_create" as well while we're at it...)

ui/urls.py
```diff
@@ -26,8 +27,17 @@
     path('media/<pk>/edit', views.MediaEditView.as_view(), name='media_item_edit'),
     path('media/<pk>', views.MediaView.as_view(), name='media_item'),
     path('channels/<pk>', views.ChannelView.as_view(), name='channel'),
+    path(
+        'create_playlist',
```
rjw57: Could we have ``playlists/create`` instead? Or, if you're worried about future primary key collision, ``playlist/create``.

frontend/src/pages/PlaylistCreatePage.js
```diff
@@ -0,0 +1,110 @@
...
+    };
+  }
+
+  /**
+   * Create the playlist.
+   */
+  create() {
+    playlistCreate(this.state.playlist)
+      .then(playlist => {
+        setMessageForNextPageLoad('The playlist has been created.');
```
rjw57: Maybe
```js
setMessageForNextPageLoad(`Playlist "${this.state.playlist.title}" created.`);
```

frontend/src/pages/MediaEditPage.js
```diff
@@ -13,16 +13,16 @@ import { setMessageForNextPageLoad } from "../containers/Snackbar";
...
+    super(props);
 
     this.state = {
       // An error object as returned by the API or the empty object if there are no errors.
       errors: {},
       // The media item being edited by the ItemMetadataForm.
       item: {},
       // remember the media item's key for convenience (read-only)
-      pk,
+      pk: props.match.params.pk,
```
rjw57: Maybe replace ``this.state.pk`` with
```js
class MediaEditPage extends Component {
  // ..
  getItemPk = () => this.props.match.params.pk;
  // ..
}
```

frontend/src/index.tsx
```diff
@@ -29,6 +30,7 @@ ReactDOM.render(
         <Route exact={true} path="/media/:pk/edit" component={MediaEditPage} />
         <Route exact={true} path="/upload" component={UploadPage} />
         <Route exact={true} path="/channels/:pk" component={ChannelPage} />
+        <Route exact={true} path="/create_playlist" component={PlaylistCreatePage} />
```
rjw57: See discussion of URL path elsewhere.

frontend/src/containers/RequiresEdit.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+  return <Component {...otherProps}>{ display }</Component>;
+};
+
+RequiresEdit.propTypes = {
+  /** A channel defining the context of the component */
+  channel: PropTypes.object,
+  /** Component to wrap children in. */
+  component: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  /** If there isn't edit permission, display this text */
```
rjw57: I think this should be an array of elements to show, like ``children``.

frontend/src/containers/RequiresEdit.js
```diff
@@ -0,0 +1,46 @@
...
+    if (channel) {
+      // if a channel is given, check that the user can edit it.
+      if (profile.channels.find((el) => (el.id === channel.id))) {
+        display = children;
+      }
+    } else {
+      // if no channel is given, check the user has channels they can edit.
+      if (profile.channels.length > 0) {
+        display = children;
+      }
```
rjw57: Since this component does two things, I think it should be two components: HiddenIfNoChannels and HiddenIfNotOwnChannel. The HiddenIfNoChannels component checks if there are any channels in the profile at all and the HiddenIfNotOwnChannel component checks if a specific channel is in the profile. They are two related, but different things. Finally, the components are checking if a profile has channels so that should be explicit in the names.

PR #242, opened 2018-08-21
Allows the user to re-order a Playlist

msb:
The user can re-order any playlist they can edit.
Partially addresses #181.
Can be reviewed commit-wise.


api/serializers.py
```diff
@@ -363,5 +363,18 @@ class PlaylistDetailSerializer(PlaylistSerializer):
 
     channel = ChannelSerializer(read_only=True)
 
+    items = serializers.SerializerMethodField(
+        help_text="Naturally ordered list of the playlist's media items")
```
rjw57: s/Naturally o/O/

api/serializers.py
```diff
@@ -363,5 +363,18 @@ class PlaylistDetailSerializer(PlaylistSerializer):
 
     channel = ChannelSerializer(read_only=True)
 
+    items = serializers.SerializerMethodField(
```
rjw57: If we have ``mediaIds``, this should probably be called ``media`` to match. Or ``mediaIds`` should be called ``itemIds``.

api/serializers.py
```diff
@@ -363,10 +363,10 @@ class PlaylistDetailSerializer(PlaylistSerializer):
 
     channel = ChannelSerializer(read_only=True)
 
-    items = serializers.SerializerMethodField(
-        help_text="Naturally ordered list of the playlist's media items")
+    media = serializers.SerializerMethodField(
+        help_text="Naturally Ordered list of the playlist's media items")
```
rjw57: I believe the review said to remove "Naturally" as well.

Issue #144, opened 2018-07-17
Investigate end-to-end testing with, e.g., selenium


rjw57:
To help with #103, investigate performing end-to-end testing using selenium. See how hard it is to add a basic "can render index" and "can render index when logged in" test is.

rjw57:
Investigated in #244.

PR #244, opened 2018-08-22
Selenium/WebDriver (WIP: DO NOT MERGE)

rjw57:
> This PR **SHOULD NOT BE MERGED**. It is a raw commit from the state of my tree at the end of the spike.

This PR documents the work I did investigating how we may integrate Selenium/Webdriver into our testing as outlined in #144. A brief summary is below followed by a list specific parts of the PR I think are of utility and ways we may wish to use them.

## Summary conclusion

**I propose that we write a stand alone end-to-end test suite which is given a URL and tests the instance of the application at that URL. This will be done as an anonymous user for the moment. The test suite can be run nightly against the test deployment on CircleCI. (This should be done in the media-deploy repo's CircleCI config.) While we're doing that, doing a nightly deploy to test would be a sensible thing to do as well.**

Although it is tempting to make end-to-end testing part of the tox test suite, I think the mechanism which is being used in this PR is brittle and prone to failure. Instead, I think it would be better to have a stand-alone end-to-end test suite which can be pointed at a deployed instance or a local development instance brought up via ``./compose.sh (development|production) up``.

## Detail

Of most interest is the new ``testutils`` application which provides a decorator for unit tests allowing them to be passed a selenium Webdriver instance for each browser we are interested in. Browserstack credentials are specified via the ``TESTUTILS_WEBDRIVER_COMMAND_EXECUTOR`` environment variable. (See the [Browserstack docs](https://www.browserstack.com/automate/python#testing-frameworks) for how to find this URL.)

In addition, a new container is added to the tox docker-compose configuration in order to allow Browserstack browsers to talk to live servers run via tox.

As a matter of plumbing, some work was done to parameterise the production Dockerfile to modify it slightly when running the test suite. Care needs to be taken here because part of the reason we want to run the tests in the production container is that we are also testing the actual container we are deploying. Having the tox container differ from production also has implications for CircleCI build times.

## Things of use

For reasons outlined above, much of the plumbing work is probably of little utility. Of greater utility is the code in the ``testutils`` module. This can be split out into a separate end-to-end test suite which can be run against a deployed instance.

## Problems

Testing a deployed instance has its own problems re authentication. Ideally we'd want to be able to test authenticated user flows as well but that has a problem in that robot accounts are currently difficult/impossible to obtain for Raven. To that end, I think we need to add a different authentication mechanism, perhaps a username/password form, which can authenticate non CRSID users. We can then add bot accounts to the deployment for testing. This has a knock-on effect for getting information on users from lookup.

At a minimum, however, there are no blockers on us writing a minimal "does it render" test for each page reachable by anonymous users.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,19 +12,27 @@ RUN npm install && npm run build
 # Use python alpine image to run webapp proper
 FROM uisautomation/django:2.1-py3.7
 
-# Ensure packages are up to date and install some useful utilities
-RUN apk update && apk add git vim postgresql-dev libffi-dev gcc musl-dev \
-	libxml2-dev libxslt-dev
+ARG tox
 
 # From now on, work in the application directory
 WORKDIR /usr/src/app
 
 # Copy Docker configuration and install any requirements. We install
 # requirements/docker.txt last to allow it to override any versions in
-# requirements/requirements.txt.
+# requirements/requirements.txt. If we're building a container for use by tox,
+# do not remove the development packages after installation, install any
+# additional tox requirements and download the local broswerstack proxy.
 ADD ./requirements/* ./requirements/
-RUN pip install --upgrade --no-cache-dir -r requirements/base.txt && \
-	pip install --upgrade --no-cache-dir -r requirements/docker.txt
+RUN apk --no-cache add git vim postgresql-dev libffi-dev gcc musl-dev \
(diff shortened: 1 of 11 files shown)


abrahammartin:
Totally agree with the conclusion. Closing the PR then but keeping the branch for future reference.

PR #247, opened 2018-08-24
mediaplatform_jwp: add mediaplatform props to created JWP videos

rjw57:
For all JWP video objects which we create, add the configured site domain and name as a custom prop. Also add the corresponding media item's id on that site. This means we can determine which media platform site a given JWP video belongs to without having to query each site's database.

After this commit, all new JWP videos will have the following custom props set:

* mediaplatform.site.name: configured human-readable name for the site
* mediaplatform.site.domain: configured domain name for the site
* mediaplatform.mediaItemId: the id of a media item on that site


rjw57:
They won't because this is only on *new* videos created on each property,
not the shared SMS videos.

rjw57:
> Isn't it called everytime that MediaItem is saved (in edits of the MediaItem title or description for example)?

Yup, but SMS media items are immutable and the only ones shared between release and so cannot be saved at the moment. All other items are created only within an individual site.

That being said, moving to a separate JWP property per deployment solves some of what this PR does so we can park it until then.

PR #249, opened 2018-08-24
DO NOT MERGE: add example navigation bar

rjw57:
> **DO NOT MERGE THIS PR.** It is intended for design discussion only.

> **UPDATE** Version 2 has been posted below.

This PR implements a sketch of how the navigation bar could look when populated with content. The content is entirely fake at the moment but illustrates some ideas for the design of the navigation bar.

screenshot_2018-08-24 the university of cambridge media platform

The PR includes code rather than just a screenshot so that the design can be tested in a variety of browsers and a variety of a screen sizes. It's intended to be the starting point for discussion much as #226 was.

### Summary of changes

1. The sign out link is reduced in visibility. This is an action that people are unlikely to want to perform so we need not take up valuable vertical screen estate with it.
2. Along with the sign out link we have links to other "secondary" navigation targets such as the "about" page, a "contact" page and a link for developers. It's important to *have* the developers' link since we should be friendly to other dev teams in the University and encourage reuse of code, etc.
3. Add a copyright notice.
4. Add a list of the user's channels. Following a link takes the user to the channel page where edit actions are available.
5. The navigation drawer has been widened by 6 spacing units. This is to allow greater room for channel names but, in accordance with the Material Design guidelines, list items do not have their heading split over multiple lines. To deal with long channel names, we "ellipsis-ise" them. This has the nice side-benefit of encouraging our users to choose snappy titles for channels.

### Rationale

It's likely that the primary interaction a user will have with the site will be to manage their channels. Each channel the user manages is available in the navigation bar. The "manage channels" destination could be an "add channel" page but navigation links should generally be nouns, not verbs.

Aside from managing channels, the other primary navigation targets for the alpha are likely to be the "help" pages and the "feedback" mechanism.

The remaining navigation destinations have been de-emphasised and placed in a lower panel. They represent navigation targets which are less important to general users.

### Other screenshots

iPhone drawer open

screen shot 2018-08-24 at 23 56 25

### DIFF
--- a/frontend/src/components/NavigationPanel.js
+++ b/frontend/src/components/NavigationPanel.js
@@ -8,11 +8,21 @@ import Button from '@material-ui/core/Button';
 import Divider from '@material-ui/core/Divider';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
+import ListItemAvatar from '@material-ui/core/ListItemAvatar';
+import ListItemIcon from '@material-ui/core/ListItemIcon';
 import ListItemText from '@material-ui/core/ListItemText';
+import ListSubheader from '@material-ui/core/ListSubheader';
 import Typography from '@material-ui/core/Typography';
 import { withStyles } from '@material-ui/core/styles';
 
+import CreateChannelIcon from '@material-ui/icons/AddToQueue';
+import ChannelIcon from '@material-ui/icons/VideoLibrary';
 import LatestMediaIcon from '@material-ui/icons/NewReleases';
+import HomeIcon from '@material-ui/icons/Home';
+import PopularIcon from '@material-ui/icons/TrendingUp';
+import FeedbackIcon from '@material-ui/icons/Feedback';
+import HelpIcon from '@material-ui/icons/Help';
+import SettingsIcon from '@material-ui/icons/Settings';
 
 /** Side panel for the current user providing navigation links. */
 const NavigationPanel = ({ profile, classes }) => <div className={ classes.root }>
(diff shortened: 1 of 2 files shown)


joeadams04:
This is nice.

Some small bits to consider:
- Manage channels should be within a channels section. This could be achieved by having a manage button next the the 'Channels' header, or by simply moving the existing link one up above the divider.
- Sign out button isn't prominent enough, it would be better situated under the account section (beneath the CRSid)
- More padding between About, Contact and Developers at the bottom

rjw57:
> Sign out button isn't prominent enough, it would be better situated under the account section (beneath the CRSid)

That's somewhat intentional. Other material approaches I've seen have made the account name + crsid a dropdown menu with a sign out option. I've not come across an example in the wild with "sign out" below the account name so I'd have to do a bit of sketching to work out how it'd live in the space.

rjw57:
OK, here's version 2. Changes:

* Sign out link is more prominent. The [Material Design Guidelines](https://material.io/design/communication/help-feedback.html#use-placement) say:
  > “Help” should be the last item in the navigation drawer, with “Send feedback” directly above it.
  >
  > Anytime [sic] “Sign out” appears in a navigation drawer, it should be the last item in the list.
* A "show more" control is added to the channel list. This is only shown if there are more than three channels. We may want to change this to "Show XX more" or "Show all XX channels".
* A "create channel" link has been added to the navigation bar in the channels section
* An "upload" link has been added in place of "popular" since we need to remember that the Media Platform is initially a publication tool, not a discovery tool.

screenshot_2018-09-01 the university of cambridge media platform

rjw57:
Closing since this is mostly implemented.

PR #253, opened 2018-08-28
Change repo name from sms-webapp to media-webapp

abrahammartin:


abrahammartin:
Auto-merging as this is required to move forward. But @rjw57 and/or @msb should review this despite been merged already as soon as they are back.

PR #255, opened 2018-08-29
Rename all the things

### LINKED ISSUE #156: Rename this repo to media-webapp and stop using sms
Will have consequences in CI, CD, etc. And documentation

abrahammartin:
This PR contain a lot of changes to many different files. It is best to review it commit base as each one of the commit does an specific thing which is described in the description of the commit.

Closes #156


mediaplatform_jwp/jwplatformapi/deliveryapi.py
```diff
@@ -16,7 +16,7 @@
 import jwt
```
msb: suggest naming the parent directory "api" as it is hard to imagine use talking a non JW api in the mediaplatform_jwp app.

mediaplatform_jwp/jwplatformapi/managementapi.py
```diff
@@ -2,10 +2,10 @@
 Interactions with the JWP management API.
```
msb: suggest calling this module management.py given api is in the parent directory

mediaplatform_jwp/jwplatformapi/deliveryapi.py
```diff
@@ -16,7 +16,7 @@
...
```
msb: suggest calling this module delivery.py given api is in the parent directory

mediaplatform_jwp/jwplatformapi/deliveryapi.py
abrahammartin: I agree

mediaplatform_jwp/jwplatformapi/managementapi.py
abrahammartin: yep it makes sense, I wanted to do it but I forgot. Initially it was called management.py but it was conflicting with the management folder/module inside mediaplatform_jwp so I changed it to managementapi.py. Now is one level down so it won't conflict.

mediaplatform_jwp/jwplatformapi/deliveryapi.py
abrahammartin: agree

PR #273, opened 2018-09-03
Drag and Drop for Mobile.

msb:
A re-factor of the playlist re-ordering code to use the "react-dnd" framework to handle drag and drop. This not only implements a superior drag and drop experience, but also implements drag and drop (although this is sub-optimal).

I'm not sure there's much value in reviewing commit-wise as I started with a reference implementation to get it working and then re-factored into cleaner components.

Closes: #274


frontend/styleguide.config.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   // propsParser: require('react-docgen-typescript').withCustomConfig('./tsconfig.json').parse,
 
   exampleMode: 'expand',
+  skipComponentsWithoutExample: true,
```
rjw57: Why was this added?

msb: So that a seperate section isn't included for DraggableContext.js as this is adequately documented in Draggable.js & Draggable.md

rjw57: So, it strikes me that ignoring components without documentation comments would be better.

Or, be explicit and use [ignore](https://github.com/styleguidist/react-styleguidist/blob/master/docs/Cookbook.md#how-to-exclude-some-components-from-style-guide).

PR #289, opened 2018-09-04
Partially implement new navigation drawer design

rjw57:
This PR partially implements the design in #249. It adds the channel list, moves the sign in/out links around and adds feedback, about, contact and developers links.

It does not implement the "create channel" link since we do not currently have that functionality in the application.

In order to provide a count of media items in the channel, this information has been added to the detail response for a channel. Since it is an aggregate statistic of all media items within a channel, this may be an expensive stat to calculate in the channel list view but it can easily be hoisted there if necessary.


api/serializers.py
```diff
@@ -385,3 +357,31 @@ class PlaylistDetailSerializer(PlaylistSerializer):
...
+    """
+    isAnonymous = serializers.BooleanField(source='user.is_anonymous')
+    username = serializers.CharField(source='user.username', required=False)
+    channels = serializers.SerializerMethodField(
+        help_text="List of channels which the user has edit rights on")
+    displayName = serializers.CharField(source='person.displayName', required=False)
+    avatarImageUrl = serializers.SerializerMethodField()
+
+    def get_channels(self, obj):
+        qs = mpmodels.Channel.objects.all().editable_by_user(obj['user'])
```
msb: as before - in the interests of having serialisers just do serialisation, would it be possible to move this onto a method on the model - or something similar?

frontend/src/components/ChannelsMuiList.js
```diff
@@ -0,0 +1,125 @@
...
+    id: PropTypes.string.isRequired,
+    mediaCount: PropTypes.number.isRequired,
+    title: PropTypes.string.isRequired,
+  })),
+};
+
+export default withStyles(styles)(ChannelsMuiList);
+
+// Used in examples
+export const mockChannels = [
```
msb: To avoid having test data in a production module, could we separate this into mockChannels.js (or similar)

PR #292, opened 2018-09-09
DO NOT MERGE: experimental re-work of media page

rjw57:
> **DO NOT MERGE THIS PR: IT IS FOR DESIGN DISCUSSION**

This PR has got a sketch of how I think the media page can be improved. We move the actions such as download, embed and show analytics to a dedicated toolbar. The description is paper which auto-collapses on small screens to connect to an app bar used to display the media information.

Add a link to the media's channel in the heading. One may also want to include a channel avatar once we have channel images sorted.

When the user can edit an item, an edit FAB is displayed.

This code is functional as far as is easy. The embed code is not yet copied to the clipboard, for example.

screen shot 2018-09-09 at 11 59 03
screen shot 2018-09-09 at 11 58 51
screen shot 2018-09-09 at 11 58 37
screen shot 2018-09-09 at 11 58 06
screen shot 2018-09-09 at 11 57 46
screen shot 2018-09-09 at 11 57 21

### DIFF
--- a/frontend/src/components/CollapsingPaperSection.js
+++ b/frontend/src/components/CollapsingPaperSection.js
@@ -1,0 +1,26 @@
+import React from 'react';
+
+import Paper from '@material-ui/core/Paper';
+import { withStyles } from '@material-ui/core/styles';
+
+const CollapsingPaperSection = ({ classes, children }) => (
+  <Paper classes={{ root: classes.paperRoot, rounded: classes.paperRounded }}>
+    { children }
+  </Paper>
+);
+
+const styles = theme => ({
+  paperRoot: {
+    [theme.breakpoints.up('md')]: {
+      margin: theme.spacing.unit,
+    },
+  },
+
+  paperRounded: {
+    [theme.breakpoints.down('sm')]: {
+      borderRadius: 0,
+    },
+  },
+});
+
+export default withStyles(styles)(CollapsingPaperSection);
--- a/frontend/src/pages/MediaPage.js
+++ b/frontend/src/pages/MediaPage.js
@@ -1,13 +1,33 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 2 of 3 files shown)


joeadams04:
On mobile I'd agree this is much better, having all the functions above the description in a small toolbar keeps it clean and compact. On desktop however I'm not so sure.

**Icon only buttons**
In the screenshots above we've lost description text when moving to icon only, which makes each action fairly ambiguous. Icon only buttons are almost always a bad idea unless used as a last resort for really obvious actions.

**Channel in header**
This is the right place to put it, however it's not obvious at all this a link to the channel and looks more like a subtitle. This is probably made worse due to the long test name of the channel, but I think it would still be hard to understand in production. I suggest adding something extra to distinguish it as a button and not plain text. The idea of a channel avatar would be great once their sorted, just make sure to provide a default, perhaps of the icon you use for channels elsewhere.

For now though you could perhaps add a border, change the background colour, add a hover style etc.  Also worth increasing opacity as this doesn't meet accessibility standards either.

**The FAB**
I think FABs perhaps have their place in mobile, however really don't work in desktop. I can explain in person the many reasons why I think this, but take a look at any new Google app (Drive, Gmail) and you'll see they're opting for different conventions for these primary actions. This is my personal opinion as FABs are obviously still part of the material guidelines.

**Paper description**
Looks odd on the desktop as it adds a clear separation between the title and channel. They're all video metadata and I think are expected to be seen together as one. If you want to separate anything it should be metadata and actions (edit, stats etc...). I think it looks fine on mobile as the paper is full width and attached to the toolbar, this too could be done to desktop to improve it.

**THE BLUE**
The blues distract way too much away from the video. It's like being in a cinema with a blue LED by your feet. Keep as much of the interface elements grayscale on the media pages.

rjw57:
Closing to have another bash later on

PR #298, opened 2018-09-11
mediaplatform_jwp: sync: do not create MediaItems for non-SMS videos

### LINKED ISSUE #297: jwpfetch creates MediaItems for all videos
jwpfetch [creates MediaItems](https://github.com/uisautomation/media-webapp/blob/master/mediaplatform_jwp/sync.py#L113) for all videos on JWP. It should only do so for videos which have an SMS media id. Otherwise, videos created via upload from media platform proliferate.

rjw57:
The jwpfetch command should not create a MediaItem for any video on JWP, only for those which were uploaded from the old SMS. Otherwise, each run of jwpfetch will create phantom media items from videos uploaded from other instances of the platform.

While we're in the mediaplatform_jwp test suite, fix a test which is not testing what it claims it is.

Closes #297


msb:
testing: the positive case passes - but when I delete a local sms video (all 4 records) to simulate a new sms video being available - the new records are created locally but without title, etc being set.

rjw57:
@msb I think I've fixed this. This whole code path is a little brittle so we should keep an eye on it as the platform grows.

msb:
testing: both the positive case and deleting a local sms video to simulate a new sms video now pass.

Issue #196, opened 2018-08-06
Show video stats in text, not only graphics


abrahammartin:

rjw57:
After some discussion, @msb and I refined this to:

1. Show total number of views from *legacy* statistics database.
2. Show total storage used on JWP.
3. Show total streamed from JWP.

PR #302, opened 2018-09-13
Displaying Additional Statistics

msb:
The additional statistics we wished to display were "total number of views", "total storage", & "total streamed". Of these, the "total streamed" was not obviously available in any of the JWP APIs. A support query has been raised with JWP and in the meantime I have implemented the display of the 2 available data points.

Can be reviewed commit wise.
Closes: #196

### DIFF
--- a/frontend/src/api.ts
+++ b/frontend/src/api.ts
@@ -43,6 +43,14 @@ const PROFILE_FROM_PAGE = (
   [0]
 );
 
+// Extract the media item's analytics if it is embedded in the page
+export const ANALYTICS_FROM_PAGE = (
+  Array.from(document.getElementsByTagName('script'))
+  .filter((element: HTMLScriptElement) => element.type === 'application/analytics+json')
+  .map((element: HTMLScriptElement) => JSON.parse(element.text))
+  [0]
+);
+
 /**
  * A function which retrieves a resource from the page by id. Note that, to avoid caching problems,
  * once retrieved, the resource is then removed from the RESOURCES_FROM_PAGE object.
--- a/frontend/src/pages/AnalyticsPage.js
+++ b/frontend/src/pages/AnalyticsPage.js
@@ -9,7 +9,7 @@ import Page from "../containers/Page";
 import Button from '@material-ui/core/Button';
 import Grid from '@material-ui/core/Grid';
 import ShowChartIcon from '@material-ui/icons/ShowChart';
-import {mediaGet} from "../api";
+import {mediaGet, ANALYTICS_FROM_PAGE} from "../api";
 import Typography from '@material-ui/core/Typography';
 
 /**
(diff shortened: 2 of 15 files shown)


frontend/src/api.ts
```diff
@@ -43,6 +43,14 @@ const PROFILE_FROM_PAGE = (
...
 );
 
+// Extract the media item's analytics if it is embedded in the page
+export const ANALYTICS_FROM_PAGE = (
+  Array.from(document.getElementsByTagName('script'))
+  .filter((element: HTMLScriptElement) => element.type === 'application/analytics+json')
+  .map((element: HTMLScriptElement) => JSON.parse(element.text))
+  [0]
+);
+
```
rjw57: I'd still like to have an aspiration of replacing all ``<a>`` tags with ``<Link>`` tags and this moves us in the wrong direction. Could this be re-factored to either a) use the same approach as ChannelView and PlaylistView which automagically fall back to using the API if the resource is not embedded in the page or b) just fetch the analytics via the API? Then we can remove the ``analytics.html`` template and this special case from ``api.ts``.

frontend/src/pages/AnalyticsPage.js
```diff
@@ -1,132 +1,153 @@
...
+  resource: mediaItem
+}) => (
+  <div>
+    <BodySection classes={{ root: classes.section }}>
+      <Typography variant="headline" component="div">
+        { mediaItem && mediaItem.title }
+      </Typography>
+    </BodySection>
+    <BodySection classes={{ root: classes.section }}>
+      <Typography variant="title" component="div" className={ classes.subtitle } >
```
rjw57: Can we use ``gutterBottom`` instead of ``classes.subtitle``?

mediaplatform_jwp/models.py
```diff
@@ -273,6 +273,15 @@ def get_sources(self):
     #: A property which calls get_sources and caches the result.
     sources = cached_property(get_sources, name='sources')
 
+    def fetch_size(self):
+        """
+        Fetches the storage size of the video in bytes (the sum of all the sources).
+        Returns 0 if unable to get a size.
+
+        """
+        cache_resource = CachedResource.videos.filter(pk=self.pk).first()
```
rjw57: I'm wondering if we want to start adding ``mediaplatform_jwp.models.Video`` and ``...Channel`` foreign keys to ``CachedResource``. (Not for this story.) Then we can stop having to roll our own queries like this.

mediaplatform_jwp/models.py
```diff
@@ -273,6 +273,15 @@ def get_sources(self):
     #: A property which calls get_sources and caches the result.
     sources = cached_property(get_sources, name='sources')
 
+    def fetch_size(self):
```
rjw57: Can we a) turn this into a ``cached_property`` called size (easy) or b) add a ``size`` field to ``mediaplatform_jwp.models.Video`` which is updated by ``jwpfetch`` (annoying)?

frontend/src/pages/AnalyticsPage.js
```diff
@@ -1,132 +1,153 @@
...
+        General Statistics
+      </Typography>
+      <Grid container>
```
rjw57: Could we have ``alignContent="center"``? I think it looks a bit weird on desktop all shoved over to the left.

mediaplatform_jwp/models.py
```diff
@@ -273,14 +273,15 @@ def get_sources(self):
     #: A property which calls get_sources and caches the result.
     sources = cached_property(get_sources, name='sources')
 
-    def fetch_size(self):
+    def get_resource(self):
```
rjw57: I think this should just be
```python
@cached_property
def resource(self):
    # ...
```
The ``sources`` property is a special case because sometimes one genuinely does want to re-call the API. With the ``resource`` prop, we're just doing database stuff so we should keep the same caching semantics as normal related fields.

frontend/src/api.ts
```diff
@@ -308,11 +319,17 @@ export const mediaUploadGet = (item: IMediaResource) : Promise<IMediaUploadResou
   return apiFetch(API_ENDPOINTS.mediaList + item.id + '/upload');
 };
 
+/** Retrieve a media resource's analytics. */
+export const analyticsGet = (id: string) : Promise<IAnalyticsResponse> => {
+  if (ANALYTICS_FROM_PAGE) { return Promise.resolve(ANALYTICS_FROM_PAGE); }
```
rjw57: Could this use ``RESOURCES_FROM_PAGE`` instead? Then one can get rid of the ``analytics.html`` template.

rjw57: We may want to change ``resourceFromPageById`` to ``resourceFromPageByPath`` and pass the path to the resource rather than an id.

Issue #296, opened 2018-09-11
Convert jwpfetch to celery task


rjw57:
Convert jwpfetch to be a celery task. Update the deployment to add a fixture which schedules it hourly.

rjw57:
Closed in #303.

PR #303, opened 2018-09-13
mediaplatform_jwp: refactor jwpfetch into celery task

rjw57:
Hoist all the jwpfetch code into a celery task and change the jwpfetch management command to call it.

This removes the need to have a separate jwpfetch cronjob; we can replace it with a scheduled task fixture in the deployment.

This is one half of #296.

### DIFF
--- a/mediaplatform_jwp/apps.py
+++ b/mediaplatform_jwp/apps.py
@@ -36,5 +36,5 @@ class Config(AppConfig):
         # Import, and thereby register, our custom signal handlers
         from . import signalhandlers  # noqa: F401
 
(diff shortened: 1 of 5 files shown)


mediaplatform_jwp/apps.py
```diff
@@ -35,3 +35,6 @@ def ready(self):
 
         # Import, and thereby register, our custom signal handlers
         from . import signalhandlers  # noqa: F401
+
+        # Import, and thereby register, out tasks
```
msb: s/b our

Issue #259, opened 2018-09-02
We need to store a backup in our HPC tape storage for every video uploaded


abrahammartin:
We should also delete them from the archive after a month of them been deleted from the app

abrahammartin:
Initially investigate how we can do it but do not close the issue

rjw57:
I think we can simply write a script to do this *á la* sms2jwplayer. *However* we should probably do this via the media-webapp API.

I've been having a look at this. For the moment, we'll ignore the existing SMS media archive and just proceed assuming we need to re-download everything.

The media API will already let a sufficiently privileged user get the list of media and then download it via the highest quality source. The key here is becoming "sufficiently" privileged.

## New permissions and permission checking

It seems the cleanest way is to make use of the [existing Django permissions system](https://docs.djangoproject.com/en/2.1/topics/auth/default/#default-permissions) coupled with the new "view" permission in Django 2.1.

Specifically:

1. We teach "viewable_by_user" to annotate "true" always if the user has the ``mediaplatform.view_mediaitem`` permission.
2. We add a new ``mediaplatform.download_mediaitem`` permission.
3. We update the source endpoint to let the ``mediaplatform.download_mediaitem`` permission trump the downloadable flag
4. We add a "super readers" group via a fixture which has the ``mediaplatform.view_mediaitem`` and ``mediaplatform.download_mediaitem`` permissions.
5. We add, either by fixture or manually, an "archive robot" user which is a member of that group.

> We may want to take the opportunity to make a similar "super editors" group and teach "editable_by_user" about the ``mediaplatform.edit_mediaitem`` permission.

We now have a robot user who can list media items ordered by created_at until it finds one it already knows about and then schedule downloads using the source download endpoint. As we have our own download endpoints which computes the JWP URL to redirect to on the fly, these URLs can be scheduled for later download since they are not time-limited.

## Authentication

Our robot user can now download all the videos but has no way of authenticating itself. We can simply turn on the DRF [TokenAuthentication](http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) backend and manually, or by fixture, [create an API token](http://www.django-rest-framework.org/api-guide/authentication/#generating-tokens) for the robot.

## Naming

[...]

PR #305, opened 2018-09-17
Support a "backup" robot user

rjw57:
> See #259 for the design discussion for this change.

In order to be able to have a robot take backup of media items, we need to add the concept of a "robot user" to the API along with any special permissions required to be able to see and download all media items.

As discussed in #259, the clean way to do this is to introduce the following features:

1. Users with the ``mediaplatform.view_mediaitem`` permission can see all media items.
2. Users with the ``mediaplatform.download_mediaitem`` permission can download all media items.
3. Users can be authenticated to the API via token as well as via the usual Django session mechanism.

This PR adds the plumbing for those features. The corresponding deploy repository will need to be updated with a user, group, permission and token fixture to allow the robot to backup from the platform.


api/serializers.py
```diff
@@ -171,6 +170,15 @@ class Meta:
...
 
+    downloadableByUser = serializers.BooleanField(
+        source='downloadable_by_user',
+        help_text=(
+            'Whether the current user can download this media item. '
+            'Some users can download media even if the downloadable flag is False.'
+        ),
+        read_only=True
+    )
+
```
rjw57: We may want to genericise this a bit into "capabilities" and have a media resource have a "capabilities" dict that looks like:
```js
{
    // ... resource ...
    "userCapabilities": {
        "download": true,
        "edit": true
    }
}
```
The "view" capability is, of course, implied by being able to see the resource :).

abrahammartin: yes, that makes it more clean

Issue #294, opened 2018-09-11
Do not attemtp to show videos in the UI if their transcode status is not "ready".


rjw57:
We have the transcode status as returned by the [Management API](https://developer.jwplayer.com/jw-platform/reference/v1/methods/videos/list.html) recorded by jwpfetch.

MediaItems whose JWP status is not "ready" should 404 when embedded and not be presented in the UI. Instead, the UI should have a message indicating that the media item has not yet finished processing or, if there was an error, any error information returned by JWP should be shown.

Expose the transcode status in the api. There should be a "status" field in the media detail response with the following values:

* "created" - Media Item was created but there is no uploaded video
* "processing" - Media Item has had a file uploaded.
* "ok" - Media Item transcode completed
* "error" - Media Item transcode errored

If "status" is "error", then the "errorDescription" field in the media detail response should contain a description of the error as returned from JWP.

Unfortunately we cannot detect at the moment when a user finishes uploading a video.

rjw57:
If #311 is merged, some of this work will be done. #311 makes sure that items whose associated JWP video's state is not "ready" are not published. As discussed in the description of that PR, the remaining work to close this issue is:

1. Create a "jwpfetch lite" task which can be run more frequently and which polls only the JWP videos we think are in the "created" or "processing" states.
2. Notify users when a video transitions to the "error" state.

PR #311, opened 2018-09-18
do not publish "unready" JWP videos

rjw57:
As noted in #294, there needs to be some better tracking of state for videos. After re-reading #294, I don't think we need to track state in the MediaItem model *just yet* since we have it available on the JWP CachedResource object.

This commit fixes an immediate pain point in that videos will be published even if they have not yet been transcoded. We update the publication check to *NOT* publish if:

1. The publication date is in the future.
2. There is a JWP video associated with the item and the video's status is not "ready".

We no longer need the NULL published_at check because or database disallows NULL publication dates (introduced in commit 809931c7).

#294 still has some work to be done on it. This change means that videos have a minimum latency of one hour to be published. We should add a task which does a "jwpfetch lite" which only checks the status of videos whose status is "created" or "processing".


abrahammartin:
as I have in on hols I may have missed something but, didn't we have a publish boolean before? If people doesn't know yet when they plan to publish their media item what are they going to enter into the field published_at (as it is not nullable)? A random date in the future doesn't seem to be a proper things to do in this case and I'd expect null values to mean that it shouldn't be publish. Thoughts?

rjw57:
We never had a published boolean.

The default for published_at is timezone.now() so if nothing is set, the video is published. There is no way, by design, to intentionally keep a video unpublished indefinitely via published_at. If you want to hide a video from everyone, this should be done via permissions.

The intent of the published_at field is to delay, not disable, publication.

abrahammartin:
unpublish video do not show up for non-editors of the video but they do for editors in the "latest media" list and the media page also works but it shows "video not found". Seems a bit of an odd behaviour.

Example admin page: https://development.media.cam.ac.uk/admin/mediaplatform/mediaitem/8-gSM4NJAWM/change/
Example "Latest media" unpublished media item showing up in the list if you can edit the media item but not showing if you are not editor: https://development.media.cam.ac.uk/
Example media item showing 404 if you are not an editor but showing the page but "video not found" embed if you are editor: https://development.media.cam.ac.uk/media/8-gSM4NJAWM

The two solutions could probably be:
a) Show differently unpublished videos for editor or not show them in the list at all
b) media page for editors should show the video instead of "video not found"

PR #318, opened 2018-09-24
Update our embed endpoint to use our own code

abrahammartin:
Instead of redirecting to the jwplayer embed code, use our own code, so we can customised it. This PR also solves the problem that the jwplayer embed url does not include google analytics. Because now we are using our own code, we can include it.

### DIFF
--- a/mediaplatform/models.py
+++ b/mediaplatform/models.py
@@ -372,17 +372,6 @@ class MediaItem(models.Model):
             ).get('size', '0')
         )
 
-    def get_jwp_embed_url(self):
-        """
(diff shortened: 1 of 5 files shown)


mediaplatform/models.py
```diff
@@ -372,15 +372,16 @@ def fetched_size(self):
             ).get('size', '0')
         )
 
-    def get_embed_url(self):
+    def get_jwp_embed_url(self):
```
rjw57: I don't think this method is actually used with the changes in api/views.py so it can probably just be removed.

PR #323, opened 2018-09-25
api: replace ListMixinBase

rjw57:
The ListMixinBase class was a little bit ugly in that it relies on MediaItem, Channel and Playlist to have exactly the same interface. It also made it hard to, e.g., only retrieve SMS objects for media items and not for playlists/channels.

Replace the mixin with a set of specialised functions which filter a queryset for a particular user and annotate any extra fields used by the serialisers. To avoid repeating ourselves too much, have a private function which implements the common functionality.

Re-factoring the ListMixinBase into a common set of functions means that it is easier to implement views. like a playlist RSS feed, which fetch related items but also respect the same permissions as the main list views.


rjw57:
Superseded by #333.

PR #326, opened 2018-09-25
Support RSS feed view of a playlist

### LINKED ISSUE #76: create an itunes U compatible RSS feed
- Example: http://rss.sms.cam.ac.uk/rss/collection/1186635?feedType=itunes and http://rss.sms.cam.ac.uk/rss/media/2657898?feedType=itunes
- This website may help into validating that is valid: https://podba.se/validate
- Requirements list at: https://help.apple.com/itc/podcasts_connect/#/itc1723472cb

Provide a RSS feed endpoint for media items and playlists. These should be protected by the view_permission field for the MediaItem or the Playlist. The RSS feed should be comply with the iTunes requirements.

RSS feeds should be protected by the view_permission field

rjw57:
> This PR depends on #322, #323, #324.

Add a specialised view of a playlist as an RSS feed available under /playlists/{pk}.rss.  The specialist view is based on the playlist API view but with a custom renderer which uses feedgen to generate the RSS feed. The feed respects the "downloadable" property and makes sure that non-downloadable videos do not appear in the list.

The Playlist serialiser was previously doing a query each time it was rendered which was inefficient. Replace this with annotating the retrieved object with a single queryset which returns all of the media items viewable by a user. This also fixes a bug whereby a playlist resource's media list could include media items invisible to the current user.

Closes #76.


rjw57:
Need to take this back into development since we need to add media item support and maybe address #332 along the way

PR #333, opened 2018-09-26
api: rework ListMixinBase

rjw57:
> This is "take 2" of #323

The ListMixinBase class was a little bit ugly in that it relies on MediaItem, Channel and Playlist to have exactly the same interface. It also made it hard to, e.g., only retrieve SMS objects for media items and not for playlists/channels.

Replace the mixin with a a mixin class designed to be used as a mixin for views, ViewMixinBase. This class contains a set of specialised methods which filter a queryset for the current user and annotate any extra fields used by the serialisers. It also adds special methods which know how to select related items used by the various detail serialisers.  To avoid repeating ourselves too much, have a private method which implements the common functionality.

This re-factoring makes it easy for views to form appropriate querysets for playlists, channels and media items. Currently this is used only to form the main view querysets but for playlists we can use the same methods to filter the returned list of media items and to make sure they are correctly annotated to be rendered by the serialisers without further database access.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -70,25 +70,94 @@ class ListPagination(pagination.CursorPagination):
     page_size = 50
 
 
-class ListMixinBase:
+class ViewMixinBase:
     """
-    Base class for the various ListView mixin classes which ensures that the list only contains
-    resources viewable by the user, annotates the resources with the viewable and editable
-    flags and selects related JWPlatform and SMS objects.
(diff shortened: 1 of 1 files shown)


abrahammartin:
as all the classes that inherit ViewMixinBase require _filter_permissions. Is't it better to put it in get_queryset?

PR #335, opened 2018-09-26
RSS feeds for media items and playlists

### LINKED ISSUE #76: create an itunes U compatible RSS feed

rjw57:
> This PR depends on #333. This PR is a re-work of #326.

Add a RSS view of media items and playlists and associated legacy redirects for RSS views of SMS media items and collections. The redirects and RSS views respect the visibility of the associated objects. Media item RSS feeds are not available for non-downloadable media items.

The media item RSS view is *only* enabled for items imported from the legacy SMS for the moment since we do not want to encourage the use of per-media item RSS feeds going forward.

The feeds are all generated with the same DRF renderer but using different serialisers for media item and playlists. The media items in the playlist feeds are filtered by visibility and downloadability.

Closes #76. Closes #326.


abrahammartin:
Testing fails in some things:
1) individual media items rss feeds do not work with rss readers/itunes. They claim that there isn't any episode.
2) Playlists seem to be broken. The API response seem to work fine but not the UI or the RSS feed (none of them showing any media item in the playlist.

rjw57:
Merged in changes from #351.

Issue #336, opened 2018-09-26
Anonymous media list page is super slow


rjw57:
The /api/media view for anonymous users is very slow. Doing some dissections it would appear that it is the ``editable_condition`` which slows things down:
```python
    def _editable_condition(self, user):
        # For the moment, we make sure that *all* SMS-derived objects are immutable to guard
        # against accidents.
        return (
            self._permission_condition('channel__edit_permission', user) &
            models.Q(sms__isnull=True) &
            models.Q(channel__sms__isnull=True)
        )
```
With the condition as above, the following ``EXPLAIN ANALYZE`` output is generated for the media list query:
```
 Limit  (cost=58088.70..58088.83 rows=51 width=429) (actual time=799.564..799.584 rows=51 loops=1)
   ->  Sort  (cost=58088.70..58089.11 rows=162 width=429) (actual time=799.564..799.580 rows=51 loops=1)
         Sort Key: mediaplatform_mediaitem.published_at DESC
         Sort Method: top-N heapsort  Memory: 55kB
         ->  Hash Left Join  (cost=18000.25..58083.30 rows=162 width=429) (actual time=228.267..779.400 rows=30217 loops=1)
               Hash Cond: ((mediaplatform_mediaitem.channel_id)::text = (mediaplatform_channel.id)::text)
               Filter: ((mediaplatform_permission.is_public AND (mediaplatform_mediaitem.published_at <= '2018-09-26 19:37:04.819162+00'::timestamp with time zone) AND ((mediaplatform_jwp_video.key IS NULL) OR ((mediaplatform_jwp_cachedresource.data -> 'status'::text) = '"ready"'::jsonb))) OR (t6.is_public AND (legacysms_mediaitem.id IS NULL) AND (legacysms_collection.id IS NULL)))
               Rows Removed by Filter: 4451
               ->  Hash Left Join  (cost=17462.49..56586.97 rows=62608 width=1970) (actual time=221.467..602.468 rows=34668 loops=1)
                     Hash Cond: ((mediaplatform_mediaitem.id)::text = (legacysms_mediaitem.item_id)::text)
                     ->  Hash Right Join  (cost=16392.55..54804.44 rows=62608 width=1942) (actual time=206.929..563.468 rows=34668 loops=1)
                           Hash Cond: ((mediaplatform_jwp_video.item_id)::text = (mediaplatform_mediaitem.id)::text)
[...]

PR #337, opened 2018-09-26
Stop anonymous media list page being quite so slow to load

rjw57:
As noted in #336, the anonymous media list page is far slower than it really should be. That issue contains some digging on the matter but the executive summary is that querying for related SMS objects in the edit condition is very slow.

The anonymous user should never have an edit permission on anything so, as suggested in #336, a quick win is to shortcut the edit condition check for anonymous users.

This does not completely fix the problem in that non-anonymous views are still slower than they should be.

Testing this change is a little challenging because the pain is only shown for larger databases; a simple GET of https://test.media.cam.ac.uk/api/media/?format=json is taking around 800ms at the moment. By contrast, on development it is around 180ms. With this PR deployed this time drops slightly to 150ms but it is likely that the "win" from this PR will only be seen when deployed to test.

### DIFF
--- a/api/tests/test_permissions.py
+++ b/api/tests/test_permissions.py
@@ -1,4 +1,6 @@
-from django.contrib.auth.models import AnonymousUser
+from unittest import mock
+
+from django.contrib.auth import get_user_model
 from django.test import TestCase
 from rest_framework.test import APIRequestFactory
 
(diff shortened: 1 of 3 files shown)


rjw57:
screenshot_2018-09-28 make it better

PR #338, opened 2018-09-27
mediaplatform_jwp: try to handle rate limit errors better

rjw57:
We have seen a lot of rate limit errors to the JWP management API in production. Attempt to be a bit smarter in retrying in the face of errors by backing off for a random duration of between 20s and 60s when we hit a rate limit. We do this 10 times before finally giving up.


rjw57:
This is currently deployed to test and I just ran a successful jwpfetch.

PR #342, opened 2018-09-28
*actually* fix 500 errors for channel view

rjw57:
> This PR supersedes #341 which contained the wrong branch :S.

Commits 5a55d421 and aedbd5a5 attempted to generalise the way that channel objects are annotated for ChannelDetailSerializer. Unfortunately, they generalised it in different ways which meant that, when merged, the generalisation was only applied to the channels rendered as part of the user profile.

Move the ProfileView over to inheriting from ViewMixinBase and teach the latter how to query the current user's profile.

This new method is used both by ProfileView and all of the views in the UI since the profile is rendered as part of each resource page. While we are refactoring, remove the erroneous non-trivial work which is done in the ResourcePageSerializer and place it in the view where it belongs.

The entire UI view/API view interaction is becoming increasingly complex and brittle. We should re-visit the entire way the UI is rendered since we don't *need* to render a corresponding resource or the user profile into the page; we can simply use the API. The only non-trivial view should be the media view since that needs to render structured JSON for SEO purposes.

To guard against the same thing happening to PlaylistView which happened to ChannelView, add a basic test.

Closes #339

PR #343, opened 2018-09-28
Rationalise UI views

rjw57:
In #342 it was noted that the tricks required to allow UI and API views to be based on the same code were becoming increasingly brittle. In fact there is no strong need for the UI to do anything particularly clever in addition to the API views except that we embed resources and user profile information into the rendered pages.

This PR is an experiment in seeing what happens if we Just Don't Do That. Aside from the media view which continues to embed JSON-LD structured data, all other views simply call on the API views to determine if a 404 should be raised or not; they all render the frontend and let the frontend fetch what it needs to via the API.

Aside from embedding resources and profiles, the only other specialisation was a custom titles. Use the NFL's excellent react-helmet library to do this client-side.

I suggest people hammer on this in a test deploy. If all seems good, this eases they way to replacing ``<a>`` tags with ``<Link>`` tags and removing the reload flash when navigating the site.


rjw57:
The flash will not be there if we move to using Link. If everyone is on board with that, then I'm happy to add Link support to the PR.

abrahammartin:
That will be great

rjw57:
@abrahammartin I've moved all the sensible uses of ``<a>`` over to ``<Link>``. This required re-working the snackbar and static pages a little which are probably the most invasive changes.

Issue #350, opened 2018-10-05
Search results page


abrahammartin:
Improve the UI of the current search results page, perhaps using a similar layout as Github which allows to select which object the user wants to see the results for.

![screenshot 2018-10-05 at 16 06 27](https://user-images.githubusercontent.com/1429691/46543406-a30ee480-c8b8-11e8-808f-41b9fb831184.png)

PR #358, opened 2018-10-12
api: add optional resource count to list endpoints

rjw57:
As sketched in #350, we need an ability to get a count of resources matching a search. This PR extends the resource list endpoints in the following ways:

  1. They now take an "include_count" parameter in the query. If "true", the response will include a total count of resource which match the other query options.
  2. They new take a "page_size" parameter which allows the page size to be customised. If the caller only wants the count of the number of resources (which they will for the search page design in #350) the page size can just be set to "1".

The page size cannot yet be limited to "0". Looking around the internets, this appears to be intentional; if the page size is zero the "next" links returned by the CursorPagination paginator become an infinite loop. This may or may not be a problem for us but, in any case, it is easy to add support for a zero page size later on if needs be.


abrahammartin:
https://development.media.cam.ac.uk/api/playlists/?include_count=true throws a 500

rjw57:
@abrahammartin I think this is related to a [Django bug](https://code.djangoproject.com/ticket/28762). I've added a (failing) test case which triggers the bug.

rjw57:
I've moved the page size functionality from this PR to #373.

PR #371, opened 2018-10-15
API: allow specifying page size

rjw57:
This PR breaks out the page size changes from #358 since they are usable on their own. We also configure a maximum page size of (arbitrarily) 300 items.


rjw57:
@msb this PR should get you cable to proceed while we wait for #358 to progress.

msb:
tested on https://test.media.cam.ac.uk for various page sizes with /media, /playlists, & /channels

Issue #102, opened 2018-07-02
Add dedicated /search route


rjw57:
We currently combine the search results with the main index page. This leads to some workarounds. Instead, support a dedicated ``/search`` route which can be re-directed to from the top-level ``AppBar`` when a search result is requested.

PR #380, opened 2018-10-18
Improved search page results

msb:
A left-hand panel is provided to navigate between the different categories of search results. The count of each of search results category is displayed in a chip next to the category selector when it becomes available.

closes: #350 
also partially addresses: #102


rjw57:
Some slight rendering issues on mobile, but happy to merge and iterate on this since the functionality is better than the current situation.

Issue #362, opened 2018-10-15
Add channel create permissions to BillingAccount


rjw57:
Add a "allows_create_channel_for_billing_account" field with related name "create_channel_permision" to Permission which links to a BillingAccount object. Create an empty Permission for each BillingAccount as a migration.

Creating a channel now requires that a billing account is specified and that the user matches the create_channel_permission on the target account.

Depends on #361.

PR #383, opened 2018-10-22
add billing account

### LINKED ISSUE #361: Add concept of a "billing account".
Add a BillingAccount object which represents a billable entity. All media platform entities are ultimately tie-able to a billing account by way of their channel.

A billing account contains the following fields:

* Lookup institution ID (required)

Move the institution from the Channel model into a corresponding billing account. All channels "owned" by the same institution id end up in the same billing account.

Required for SMS switch off because we want to have multiple channels per billable entity.

rjw57:
Add the concept of a billing account to the application. A billing account is associated with each channel and is intended to both control channel creation and to allow tracing resources back to a billable entity.

This PR includes some non-trivial migrations which attempt to preserve the association of channels and institutions both forwards an backwards across migrations. I suggest rocking the DB from migration 0020 to 0022 nd back a few times to convince yourself of this.

The API is updated to know about billing accounts and to be able to list them. For the moment the list of billing accounts any user can add channels to is empty which means that channel creation is effectively disabled. The implementation of #362 will need to add a permissions model around billing accounts.

Much of the donkey work in the PR is fixing tests to know that channels must now have a billing account.

While we're modifying the database schema, this PR also takes the time to add indexes for the remaining '..._at' fields which lacked them and also performs an automated migration which Django is complaining about at the moment.

Closes #361


api/views.py
```diff
@@ -125,6 +125,17 @@ def filter_playlist_qs(self, qs):
         # For the moment we only need to respect permissions
         return self._filter_permissions(qs)
 
+    def filter_billing_account_qs(self, qs):
+        """
+
```
msb: missing method description

api/views.py
```diff
@@ -158,6 +169,15 @@ def add_playlist_detail(self, qs):
         """
         return qs.select_related('channel')
 
+    def add_billing_account_item_detail(self, qs):
```
msb: In the spirit of YAGNI, my preference (not strong) would be to add this when you need it.

abrahammartin:
Added issue about django reversion https://github.com/uisautomation/media-webapp/issues/384

PR #385, opened 2018-10-29
frontend: port SearchFrom typescript component from pidash

rjw57:
The pidash is typescript-only and nicely ported the SearchForm component for us. Acknowledge the favour and take the SearchForm port back from it.


msb:
Tested on https://development.media.cam.ac.uk.

PR #386, opened 2018-10-30
Add billing account channel create permissions

### LINKED ISSUE #362: Add channel create permissions to BillingAccount

rjw57:
Add a permission to BillingAccount which determines if a user can create channels associated with the account. Expose this in the API as requiring that the billiingAccountId parameter be set when creating channels. Add the ability for users to determine which accounts they can create channels on by way of a custom filter on the billing accounts list endpoint.

Closes #362.


msb:
Tested on https://development.media.cam.ac.uk - checked that channels could / couldn't be created for billing accounts with / without the allows_create_channel permission, respectively.

PR #389, opened 2018-11-02
mediaplatform_jwp: use sms_collection_media_ids custom prop

### LINKED ISSUE #388: Playlist sync appears broken
See, e.g., https://alpha.media.cam.ac.uk/playlists/csulMCINGbA and https://alpha.media.cam.ac.uk/channels/BUN98cUsgnw. These should have media items, e.g., https://alpha.media.cam.ac.uk/media/1mbtYg6a4Fw but they do not.

Looking in the admin, the playlist is empty and the channel is unassigned for that video making me think that we have a jwpfetch or an sms2jwplayer issue. The JWP playlist is empty leading me to think the sms2jwplayer script is broken.

rjw57:
Playlist and channel synchronisation was making use of the sms_media_ids custom prop which is only set once sms2jwplayer has managed to fully synchronise the playlist. We don't actually care any more about which videos are in the playlist so commit adbc0ee in sms2jwplayer added a new prop, sms_collection_media_ids, which list which ids should be in a collection and in which order.

Move over to using that prop instead of sms_media_ids.

Closes #388

### DIFF
--- a/mediaplatform_jwp/sync.py
+++ b/mediaplatform_jwp/sync.py
@@ -363,17 +363,29 @@ def update_related_models_from_cache(update_all_videos=False):
 
         channel.edit_permission.save()
 
-        # Update contents
+        # Update contents. We use the "sms_collection_media_ids" custom prop as that is always set
+        # to the media ids which "should" be in the collection unlike sms_{,failed_}media_ids which
+        # is used as part of the playlist synchronisation process.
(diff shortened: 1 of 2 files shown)


rjw57:
Deployed to test and confirmed to fix #388. Deploying as a matter of urgency because it is a user-visible bug.

PR #391, opened 2018-11-16
doc: add a brief developer guide on uploading media items

rjw57:
The guide is fairly brief since there's not much to say. I've not actually tested trying to upload something using just the information in the guide since I think I'm too familiar with the MP to not use external knowledge. It would be useful for someone unfamilar with the platform to review the guide.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -28,3 +28,7 @@ indent_size=2
 indent_style=space
 indent_size=2
 max_line_length=99
+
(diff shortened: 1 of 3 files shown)


abrahammartin:
All good to me, I'd only add the creation of a playlist to the docs as this is necessary to implement an Opencast workflow.

rjw57:
Good idea, done.

PR #392, opened 2018-11-19
ensure media upload endpoints are one-off

### LINKED ISSUE #367: Upload endpoints should be one-off.
Upload endpoints are single use and we don't know when a user uses them so make sure that a given endpoint is only returned once from the API. The upload API endpoint should be POST-only. POST-ing to it returns the upload endpoint if one exists or creates and returns one if one doesn't. If an UploadEndpoint exists and the POST returns immediately, it should delete the UploadEndpoint. If an UploadEndpoint does not exist, the POST view should create an ephemeral one which is returned but not persisted to the DB.

UploadEndpoints should not be retrieved for videos whose status is anything other than "created". If the video has a different status, the endpoint should return an error response that indicates that a video has already been uploaded.

The idea is that we get an UploadEndpoint creation "for free" when we create the media item at the moment. We want the common case the an API user creates a media item and then retrieves the upload endpoint to be cheap requiring only one JWP management API call but we want to support re-tries for uploading the video.

rjw57:
JWP media upload endpoints are one-use-only URLs. Unfortunately we allowed the same URL to be retrieved multiple times from the upload endpoint which meant that a client may try to upload a media item multiple times to the same URL.

Fix this by making the /media/{id}/upload endpoint a PUT-only endpoint.  If the upload endpoint URL exists in the database to begin with, we return it and delete it from the database. If the endpoint does not exist, we create one first.

This involves a bit of an unnecessary write and then delete from the database in the case where the endpoint does not exist but for the common case of "create media item" then "get upload endpoint", it Does The Right Thing (TM).

Closes #367


api/serializers.py
```diff
@@ -309,8 +311,31 @@ class MediaUploadSerializer(serializers.Serializer):
...
+        """
+        Ensure that the instance has an upload endpoint which has not expired. Delete it just
+        before returning it to the rest of the serialiser so that we are sure we only ever give an
+        upload URL back to a client once.
+
+        """
+        # If there is already a created upload endpoint which expires more than a day from now,
+        # we can use the instance as is.
+        if hasattr(instance, 'upload_endpoint'):
+            headroom = datetime.timedelta(days=1)
```
abrahammartin: will this code ever get executed if the we always delete the instance from the database after we create it?

rjw57: Yes as an upload endpoint is created by JWP when we first create the media resource. See the last line of ``mediaplatform_jwp.api.management._perform_item_update()``.

abrahammartin (approved):

abrahammartin (changes requested):
Executing a second PUT to the uploads API makes it fail. The error raised by the JWPlayer API is: JWPlatformPreconditionFailedError
'Video must have status `ready` or `failed` before its original file can be updated. Current status is: processing.'

rjw57:
If the PUT succeeds after JWP has finished processing, I'd be inclined to call that good enough to fix the issue for the moment. We're likely to migrate away from JWP so re-woking all the API plumbing to support it is probably left for another issue.

PR #399, opened 2018-11-27
Do not show media invisible to user in playlist view

### LINKED ISSUE #398: Playlist detail view on API does not resepect media view permissions
E.g. https://test.media.cam.ac.uk/api/playlists/Vuva7ojixto - shows items which the anonymous user should not be able to see.

rjw57:
As noted in #398, we were inadvertently showing all media in playlists irrespective of whether the user should be able to see them. This PR consists of two commits; one closes the issue and the other adds a test for media items being returned in playlists which was missing.

Closes #398.

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -411,7 +411,7 @@ class PlaylistDetailSerializer(PlaylistSerializer):
 
     channel = ChannelSerializer(read_only=True)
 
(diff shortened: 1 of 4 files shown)


msb:
Tested locally - checked that an item in a playlist could be seen before but not after the change

PR #400, opened 2018-11-27
Make track types configurable

rjw57:
Previously we hard-coded the track types which would be promoted to media items as a setting. This PR keeps that setting but uses it as a default to populate a field in Series objects. That field specifies which tracks should be uploaded as media items and can be changed per Series.

This PR also improves the list display for Series in the admin to include the repository which is useful when there are multiple repositories being imported.


rjw57:
I'm going to deploy to dev for testing.

rjw57:
Closing since I think we have a better way to do it.

PR #406, opened 2018-12-10
Explicitly included the 'fetch' polyfill.

msb:
Explicitly included the 'fetch' polyfill to potentially avoid the IE11 'fetch not found' bug experienced with the Research Dashboard.


rjw57:
Tested in the development instance.

PR #409, opened 2018-12-12
mediaplatform_jwp: use correct URL signing for player.js URLs

### LINKED ISSUE #407: URL signing mechanism is incorrect for JS player URL
The overnight test deployment flagged up that we are signing the ``/libraries/player.js`` URL incorrectly. It appears that URLs for the [delivery API](https://developer.jwplayer.com/jw-platform/docs/delivery-api-reference/) do not all use JWT signing.

This was not noticed in development because we had not enabled the requirement that all URLs be signed in the development JWP property.

We need to make use of the "[legacy token signing](https://developer.jwplayer.com/jw-platform/docs/developer-guide/delivery-api/legacy-url-token-signing/)" mechanism for JS players.

Required URL signing has now been enabled for the development property and so running the platform locally is currently broken.

### LINKED ISSUE #408: Searching on embedded playlists breaks selecting videos
When searching on a playlist view, the non-matching videos are hidden but clicking on a video uses the index of the video in the filtered lists to select the video to play rather than the index in the unfiltered list.

rjw57:
As noted in #407, we were signing player.js URLs incorrectly. Use the legacy signing method for player URLs as documented at [1].

[1] https://developer.jwplayer.com/jw-platform/docs/developer-guide/delivery-api/legacy-url-token-signing/

Closes #407

### DIFF
--- a/mediaplatform_jwp/api/delivery.py
+++ b/mediaplatform_jwp/api/delivery.py
@@ -448,6 +448,58 @@ def pd_api_url(resource, now_timestamp=None, **parameters):
     return url
 
 
+def legacy_pd_api_url(resource, now_timestamp=None, **parameters):
+    """
+    Return a JWPlatform Platform Delivery API URL with the request has the appropriate signature
+    and expiry parameters attached.
+
+    :param str resource: path to resource, e.g. ``/libraries/abcdef.js``. Must start with a slash.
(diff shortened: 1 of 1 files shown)


rjw57:
Pulling as signing mechanism is insecure. We'll proxy the player instead.

PR #411, opened 2018-12-12
mediaplatform_jwp: proxy player library

### LINKED ISSUE #407: URL signing mechanism is incorrect for JS player URL

rjw57:
As noted in #407, we were signing player.js URLs incorrectly. Implement the legacy signing method for player URLs as documented at [1]. Looking at the signing method, there are a number of weaknesses:

1. MD5 is a broken hash.

2. The secret is appended to the value to be hashed meaning that the hash state preceding the secret is known. HMAC, for example, guards against this explicitly.

To avoid exposing the signatures to all and sundry, we proxy the player library so that the signature only goes between our server and JWP. We cache the resulting response to avoid making a round-trip delay to JWP be the common case and instead to make the common case that the player JavaScript is served from the cache.

[1] https://developer.jwplayer.com/jw-platform/docs/developer-guide/delivery-api/legacy-url-token-signing/

Closes #407

### DIFF
--- a/ui/views.py
+++ b/ui/views.py
@@ -4,8 +4,9 @@ Views
 """
 import logging
 
-from django.http import Http404
-from django.views.generic.base import RedirectView
+from django.http import Http404, HttpResponse
+from django.views.decorators.cache import cache_page
+import requests
 from rest_framework import generics
 from rest_framework.renderers import TemplateHTMLRenderer
 from rest_framework.serializers import Serializer as NullSerializer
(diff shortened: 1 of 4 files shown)


ui/views.py
```diff
@@ -142,11 +143,17 @@ def get_object(self):
         return playlist
 
 
-class PlayerLibraryView(RedirectView):
+# We cache this view for 15 minutes since it introduces a round trip JWP API latency for each
+# uncached request.
+@cache_page(60 * 15)
```
abrahammartin: good thinking

abrahammartin:
tested and working

PR #420, opened 2019-01-07
mediaplatform: new model: TranscriptionRequest

### LINKED ISSUE #403: LC8: Include a subtitles flag in media items
We want to include a flag in the MediaItem class that would indicate us if we need to generate subtitles (audio transcriptions) for that media item. ~~This will be on by default for some pre-selected lecture capture series.~~

rjw57:
Add a new model to the mediaplatform application which records a transcription request and the state of the request. Add some basic configuration to the admin so that requests may be searched, listed and created.

Closes #403


mediaplatform/models.py
```diff
@@ -896,6 +896,70 @@ def __str__(self):
...
+    """
+    A transcription request represents a request that a media item have a caption track appended. A
+    request may be in one of five states.
+
+    * A "pending" request is one which is outstanding and may be picked up by some transcription
+        service.
+    * A "running" request is one which is currently underway.
+    * A "resolved" request is one which finished and resulted in a caption track.
+    * A "rejected" request is one which finished and resulted in some problem.
+    * A "cancelled" request is one which should not be picked up by a transcription service but was
```
abrahammartin: confusing comment

